PR #3, opened 2025-02-25
chore: remove `SetTheory` namespace

vihdzp:
This isn't really necessary anymore now that all of this has its own repo.

PR #4, opened 2025-02-25
chore: move "special" games to their own file

vihdzp:
- [x] depends on: #3


README.md
```diff
@@ -1,3 +1,18 @@
 # Combinatorial games in Lean
 
-TODO: write a proper description!
\ No newline at end of file
+A formalization of topics within combinatorial game theory in Lean 4.
```
tristan-f-r: (this diff shouldn't be in this PR, though I suppose that will be automatically resolved when this PR is merged)

CombinatorialGames/Game/Specific.lean
```diff
@@ -0,0 +1,130 @@
+/-
+Copyright (c) 2022 Violeta Hernández Palacios. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Violeta Hernández Palacios, Tristan Figueroa Reid
+-/
+import CombinatorialGames.Game.PGame
+
+/-!
+# Specific games
```
tristan-f-r: I think 'Specific' is a bad word here, because other 'specific' games defined in the literature are sometimes modifiers on games (i.e. Tiny & Miny games): perhaps 'Special' is a better word, since these types of games usually have interesting resulting theorems?

CombinatorialGames/Game/Birthday.lean
```diff
@@ -5,6 +5,7 @@ Authors: Violeta Hernández Palacios
 -/
 import Mathlib.Algebra.Order.Group.OrderIso
 import CombinatorialGames.Game.Ordinal
+import CombinatorialGames.Game.Specific
```
tristan-f-r: Should theorems about birthdays/impartialness be moved to Specific?

CombinatorialGames/Game/Specific.lean
tristan-f-r: (Or, we can have a separate file for those special games/game modifiers)

LeoDog896 (changes requested):

README.md
vihdzp: Whoops, I must've pushed that to the wrong branch.

CombinatorialGames/Game/Specific.lean
vihdzp: I wasn't aware that "specific" games were already a thing! I think we can keep calling them "special", then.

CombinatorialGames/Game/Birthday.lean
vihdzp: I think this file should preferably as low as possible in the import chain. We should be able to import special games in `Impartial` / `Birthday` / all those other files and prove whatever the relevant properties are, not the other way around, unless we want these simple games to be bloated behind a bloated file.

LeoDog896 (approved):

PR #8, opened 2025-02-25
feat: improve iff lemmas on impartial games

vihdzp:
Mathlib PR [#21110](https://github.com/leanprover-community/mathlib4/pull/21110).


vihdzp:
Whoops, forgot we weren't doing deprecations for now! I've removed those from the PR.

PR #7, opened 2025-02-25
feat: strategy stealing

vihdzp:
Mathlib PR [#21073](https://github.com/leanprover-community/mathlib4/pull/21073).

PR #9, opened 2025-02-25
feat: poset games

vihdzp:
Mathlib PR [#21003](https://github.com/leanprover-community/mathlib4/pull/21003).

- [x] depends on: #3
- [x] depends on: #7
- [x] depends on: #14


LeoDog896 (changes requested):
I couldn't golf any part of this file, and the definitions seem good.

My only problem with this file conceptually is that it would be useful to have non-impartial poset games (i.e. while Green Hackenbush is a special impartial poset game, Red-Green-Blue Hackenbush is a special partisan poset game).

LeoDog896 (approved):

PR #13, opened 2025-02-25
feat: add README

vihdzp:


vihdzp:
Loopy games are definitely something we want, though perhaps we'll have to think a bit about how to implement it within our framework. We can update the readme after the fact.

PR #14, opened 2025-02-26
feat: concrete games

vihdzp:
We define a concrete game as a type endowed with two well-founded relations, determining the ways in which Left and Right can change the game state. We then provide some basic API to turn a concrete game state into a `PGame`.

This is intended to supersede the existing `State` API, which only works for finite games.

### DIFF
--- a/CombinatorialGames/Game/Concrete.lean
+++ b/CombinatorialGames/Game/Concrete.lean
@@ -79,6 +79,7 @@ Even though these types are the same (not definitionally so), this is the prefer
 between them. -/
 def toLeftMovesPGame {a : α} : {b // b ≺ₗ a} ≃ (toPGame a).LeftMoves :=
   Equiv.cast (leftMoves_toPGame a).symm
+
 /-- Turns a right-subsequent position for `a` into a right move for `toPGame a` and vice versa.
 
 Even though these types are the same (not definitionally so), this is the preferred way to convert
--- a/CombinatorialGames/Game/Concrete.lean
+++ b/CombinatorialGames/Game/Concrete.lean
@@ -119,7 +119,7 @@ theorem toRightMovesPGame_symm_prop {a : α} (i : (toPGame a).RightMoves) :
 theorem neg_toPGame (h : subsequentL (α := α) = subsequentR) (a : α) : -toPGame a = toPGame a := by
   rw [toPGame, neg_def]
   congr
-  all_goals try (ext; rw [h])
+  any_goals ext; rw [h]
   all_goals
     apply Function.hfunext (by rw [h])
     simp_rw [heq_eq_eq, Subtype.forall, h]
--- a/CombinatorialGames/Game/Concrete.lean
+++ b/CombinatorialGames/Game/Concrete.lean
@@ -37,14 +37,13 @@ attribute [instance] isWellFounded_subsequent
 theorem subrelation_subsequentL :
     Subrelation subsequentL fun a b : α ↦ subsequentL a b ∨ subsequentR a b :=
   Or.inl
+
 theorem subrelation_subsequentR :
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Game/Concrete.lean
```diff
@@ -0,0 +1,146 @@
...
+    apply Function.hfunext (by rw [h])
+    simp_rw [heq_eq_eq, Subtype.forall, h]
+    intro a ha b hb he
+    have : a = b := heq_subtype (by rw [h]) he
+    subst this
+    have := subrelation_subsequentR hb
+    apply neg_toPGame h
+termination_by isWellFounded_subsequent.wf.wrap a
+
+theorem impartial_toPGame (h : subsequentL (α := α) = subsequentR) (a : α) :
```
vihdzp, line 133: Should this theorem be here, or in the impartial file?

CombinatorialGames/Game/Concrete.lean
```diff
@@ -0,0 +1,141 @@
...
+variable [ConcreteGame α]
+
+scoped infix:50 " ≺ₗ " => subsequentL
+scoped infix:50 " ≺ᵣ " => subsequentR
+attribute [instance] isWellFounded_subsequent
+
+theorem subrelation_subsequentL :
+    Subrelation subsequentL fun a b : α ↦ subsequentL a b ∨ subsequentR a b :=
+  Or.inl
+theorem subrelation_subsequentR :
```
tristan-f-r, line 41: I am generally made uncomfortable by the difference in newline styling compared to mathlib4 (zero newlines for related theorems and one newline for slightly related theorems, compared to one newline and two newlines respectfully), though I suppose that is okay.

CombinatorialGames/Game/Concrete.lean
LeoDog896: The theorem being here seems fine: I can see many impartial games being defined, and this theorem is small enough that it doesn't warrant some ConcreteImpartial file.

CombinatorialGames/Game/Concrete.lean
```diff
@@ -0,0 +1,141 @@
...
+
+/-- Use `toLeftMovesPGame` to cast between these two types. -/
+theorem leftMoves_toPGame (a : α) : (toPGame a).LeftMoves = {b // b ≺ₗ a} := by
+  rw [toPGame]; rfl
+/-- Use `toRightMovesPGame` to cast between these two types. -/
+theorem rightMoves_toPGame (a : α) : (toPGame a).RightMoves = {b // b ≺ᵣ a} := by
+  rw [toPGame]; rfl
+
+theorem moveLeft_toPGame_hEq (a : α) :
+    HEq (toPGame a).moveLeft fun i : {b // b ≺ₗ a} => toPGame i.1 := by
```
tristan-f-r: Good use of HEq here!

LeoDog896 (approved):

CombinatorialGames/Game/Concrete.lean
vihdzp: I've seen some limited usage of "theorems without whitespace in between" in Mathlib, though it's often with one-liners and maybe I'm pushing the envelope a bit too hard. I'll space out these longer theorems.

CombinatorialGames/Game/Concrete.lean
vihdzp: I copied this hack over from the nim code; I've got no idea how I came up with it back then.

CombinatorialGames/Game/Concrete.lean
tristan-f-r: It can stay here: Impartial games are common enough that this theorem can stay in Concrete for now, though I could see a future where `ImpartialConcrete` is realized.

PR #18, opened 2025-02-26
refactor: use `ConcreteGame` in domineering

vihdzp:
- [x] depends on: #14

Some more battle testing for this new API.


CombinatorialGames/Game/Specific/Domineering.lean
```diff
@@ -0,0 +1,155 @@
...
+Specifically to domineering, we need the fact that
+disjoint parts of the chessboard give sums of games.
+-/
+
+namespace PGame
+
+open Function
+
+private theorem pred_fst_ne (m : ℤ × ℤ) : (m.1 - 1, m.2) ≠ m := by aesop
+private theorem pred_snd_ne (m : ℤ × ℤ) : (m.1, m.2 - 1) ≠ m := by aesop
```
tristan-f-r:
```suggestion
-- Note: `aesop` can't prove pred_*_ne when the inequality is reversed
private theorem pred_fst_ne (m : ℤ × ℤ) : (m.1 - 1, m.2) ≠ m := by aesop
private theorem pred_snd_ne (m : ℤ × ℤ) : (m.1, m.2 - 1) ≠ m := by aesop
```

CombinatorialGames/Game/Specific/Domineering.lean
```diff
@@ -0,0 +1,155 @@
...
+
+/-- A Domineering board is an arbitrary finite subset of `ℤ × ℤ`. -/
+def Domineering := Finset (ℤ × ℤ) deriving DecidableEq
```
tristan-f-r, line 30: (I would like to add `Repr` here, though it doesn't seem `mathlib4` has Repr on `(ℤ × ℤ)`?)

CombinatorialGames/Game/Specific/Domineering.lean
tristan-f-r: (I also would like to be more sparing of the use of `aesop`, and this would probably be the only place where I would suggest using `have ...; simp` instead of `aesop`, but since this is a private theorem, this makes sense :+1:).

vihdzp: Just realized I only ended up using these theorems once... I think I can just inline them and replace them by a short term-mode proof.

vihdzp: I do think the other places in the file where I used `aesop` are quite appropriate uses of it, e.g. the theorem
```
(ofDomineering (toDomineering (((ofDomineering b).erase m).erase (m.1 - 1, m.2)))).card + 2 = (ofDomineering b).card
```
is "structurally trivial" but would take a couple of lines of code to prove manually.

CombinatorialGames/Game/Specific/Domineering.lean
vihdzp: The problem isn't `ℤ × ℤ`, but `Finset`. There's an unsafe `Repr` instance in a file that wasn't previously imported - I've used it to construct the `Repr Domineering` instance.

LeoDog896 (approved):

CombinatorialGames/Game/Specific/Domineering.lean
```diff
@@ -21,154 +24,133 @@ disjoint parts of the chessboard give sums of games.
...
+  toDomineering (ofDomineering a) = a := rfl
+@[simp] theorem ofDomineering_toDomineering (s : Finset (ℤ × ℤ)) :
+  ofDomineering (toDomineering s) = s := rfl
 
-/-- A Domineering board is an arbitrary finite subset of `ℤ × ℤ`. -/
-abbrev Board :=
-  Finset (ℤ × ℤ)
+namespace Domineering
+
+unsafe instance : Repr Domineering := Finset.instRepr
```
tristan-f-r, line 41: Right! I forgot that `Finset` order is locked under `unsafe`.

vihdzp: We could use `Finset.sort` to give a safe instance. But I don't think that's really necessary - when are we going to want to prove things about `Repr` anyways?

PR #19, opened 2025-02-26
chore: add vscode ruler

LeoDog896:
artificially imposes the 100-character line limit in vscode with a visual ruler.


vihdzp:
Nice! This is definitely worth if we're sticking to general Mathlib style guidelines.

vihdzp (approved):

PR #20, opened 2025-02-26
chore: replace `fun _ => _` by `fun _ ↦ _`

vihdzp:
This would be seen as complete bikeshedding over at Mathlib... but since the codebase is so much smaller now I think we can afford some stylistic regularity.

Did a few other minor golfs where I was able to find them.


LeoDog896 (approved):
I stylistically agree with `↦`.

All golfs and cleanups make sense :+1:

PR #25, opened 2025-02-26
feat: add various recursors

vihdzp:
We previously had a few theorems for casing on moves of more complicated operations. We promote these to `Sort*`-valued recursors, and prove that they behave as they should.


vihdzp:
Closing in favor of the `IGame` refactor.

PR #26, opened 2025-02-26
doc: document type equalities

vihdzp:
There's a bunch of theorems stating equalities between types that you're *usually* not supposed to use directly. We add comments explaining what casting operation to use instead.

For those rare cases where you *do* want to use these equalities directly (e.g. proving a `Nonempty` instance), we tag some `simp` lemmas to facilitate this. (These `simp` lemmas will never fire when the types are used dependently, which should hopefully mean this won't break things later down the line).

Also, some drive-by compression.

### DIFF
--- a/CombinatorialGames/Game/PGame.lean
+++ b/CombinatorialGames/Game/PGame.lean
@@ -132,21 +132,10 @@ def moveLeft : ∀ g : PGame, LeftMoves g → PGame
 def moveRight : ∀ g : PGame, RightMoves g → PGame
   | mk _ _r _ R => R
 
-@[simp]
-theorem leftMoves_mk {xl xr xL xR} : (⟨xl, xr, xL, xR⟩ : PGame).LeftMoves = xl :=
-  rfl
-
-@[simp]
-theorem moveLeft_mk {xl xr xL xR} : (⟨xl, xr, xL, xR⟩ : PGame).moveLeft = xL :=
-  rfl
-
-@[simp]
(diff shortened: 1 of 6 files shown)


CombinatorialGames/Game/PGame.lean
```diff
@@ -162,41 +151,33 @@ lemma ext {x y : PGame} (hl : x.LeftMoves = y.LeftMoves) (hr : x.RightMoves = y.
...
 
+/-- Use `toOfListsRightMoves` to cast between these two types. -/
+@[simp]
 theorem rightMoves_ofLists (L R : List PGame) : (ofLists L R).RightMoves = ULift (Fin R.length) :=
   rfl
 
-/-- Converts a number into a left move for `ofLists`.
-
-This is just an abbreviation for `Equiv.ulift.symm` -/
-abbrev toOfListsLeftMoves {L R : List PGame} : Fin L.length ≃ (ofLists L R).LeftMoves :=
```
tristan-f-r, line 176: Did you mean to drop these abbrves when using referencing them in `left/rightMoves_ofLifsts`?

tristan-f-r (changes requested):

CombinatorialGames/Game/PGame.lean
vihdzp: These abbrevs aren't very useful, since the relevant type equality is definitional. That said... I think it's better to just close this PR. Presumably the plan is to port everything to `IGame`, and I think I've shown that `IGame` has no need for these type rewriting hacks.

PR #28, opened 2025-02-26
feat: add recursors for `insertLeft` and `insertRight`

vihdzp:
We use these to golf existing results.

We also move all of this to its own file.


vihdzp:
Closing in favor of the `IGame` refactor.

Issue #29, opened 2025-02-27
Move things over to `IGame`


vihdzp:
In mathematics, combinatorial games are often defined within ZFC, in the framework of sets. An implicit yet important consequence of this is that games satisfy the axiom of extensionality: if `G` and `H` have the same left/right moves, then they are the same game.

Our development of games has historically not satisfied this axiom. By instead defining "base" games `PGame` inductively as consisting of indexed families families over itself, we can only ever have two games be equal when they're indexed by the *same* functions; in particular, by the *same* types. So even simple equalities like { | } = 0 are unprovable (for all but one choice of empty types to index the left and right games over), and others such as {0, 1, 1 | } = 2 are actually just false. Though the quotients `Game` and `Surreal` do match their usual mathematical definitions, every time we need to go through `PGame` it's just painful.

The proposal here is to relegate `PGame` to an auxiliary definition, used to set up the real type we actually care about: `IGame`, i.e. the quotient of `PGame` by extensionality. Though limitations on inductive types mean we can't define it this way, `IGame` can be recursively built through:
```lean 4
def ofSets (s t : Set IGame.{u}) [Small.{u} s] [Small.{u} t] : IGame.{u} := ...

def ofSetsRecOn {P : IGame.{u} → Sort*} (x)
    (H : Π (s t : Set _) [Small s] [Small t], (Π x ∈ s, P x) → (Π x ∈ t, P x) → P (ofSets s t)) : P x :=
  ...
```
Which means, once we prove this, we have no reason to use `PGame` anymore.

This refactor will probably set us back a few weeks, since we'll be moving basically all of the code that had already been written to this new framework. But I do expect that as a result on this, proving new results will be much easier, and the results will be more aligned with the traditional ZFC theory.

PR #30, opened 2025-02-27
feat: `IGame` part 1

vihdzp:
We define up to the `AddCommMonoid` instance on `IGame`. #29

There's some results I haven't ported yet like `x - x ≈ 0` or `x ≤ y → z + x ≤ z + y`. I think a lot of the results that existed on `PGame` could be simplified by first building the quotient `Game = Antisymmetrization IGame`, proving the `OrderedAddCommGroup` instance on it, and then "porting back" the results via def-eq. So I'd like to do that before proving the more advanced relations between orders and arithemetic on `IGame`.

For now I've put everything in an `IGame` folder - of course, when the refactor is over, the idea is to put everything back in its place, overwriting whatever we had before.

### DIFF
--- a/CombinatorialGames/IGame/IGame.lean
+++ b/CombinatorialGames/IGame/IGame.lean
@@ -19,17 +19,15 @@ open PGame Set Pointwise
 
 /-- Games up to identity.
 
-`IGame` uses the set-theoretic notion of equality on Games,
-compared to `PGame`'s 'type-theoretic' notion of equality.
+`IGame` uses the set-theoretic notion of equality on games, compared to `PGame`'s 'type-theoretic'
+notion of equality.
 
-This is not the same equivalence as used broadly in combinatorial game theory literature,
-as a game like {0,1|0} is not _identical_ to {1|0}, despite being equivalent.
-However, many theorems can be proven over the 'identical' equivalence relation,
-and the literature may occasionally specifically use the 'identical' equivalence
-relation for this reason. 
+This is not the same equivalence as used broadly in combinatorial game theory literature, as a game
+like {0, 1 | 0} is not *identical* to {1 | 0}, despite being equivalent. However, many theorems can
+be proven over the 'identical' equivalence relation, and the literature may occasionally
+specifically use the 'identical' equivalence relation for this reason.
 
-For game equivalence from literature, see `Game.Basic`.
--/
+For the more common game equivalence from literature, see `Game.Basic`. -/
 def IGame : Type (u + 1) :=
   Quotient identicalSetoid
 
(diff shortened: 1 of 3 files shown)


CombinatorialGames/IGame/IGame.lean
```diff
@@ -0,0 +1,529 @@
...
+-- This is a false positive due to the provisional duplicated IGame/IGame file path.
+set_option linter.dupNamespace false
+-- All computation should be done through `IGame.Short`.
+noncomputable section
+
+open PGame Set Pointwise
+
+/-- Games up to identity.
+
+TODO: write proper docstring. -/
```
tristan-f-r:
```suggestion
`IGame` uses the set-theoretic notion of equality on Games,
compared to `PGame`'s 'type-theoretic' notion of equality.

This is not the same equivalence as used broadly in combinatorial game theory literature,
as a game like {0,1|0} is not _identical_ to {1|0}, despite being equivalent.
However, many theorems can be proven over the 'identical' equivalence relation,
and the literature may occasionally specifically use the 'identical' equivalence
relation for this reason. 

For game equivalence from literature, see `Game.Basic`.
-/
```

CombinatorialGames/IGame/IGame.lean
```diff
@@ -0,0 +1,529 @@
...
+  isOption_wf.recursion x fun x IH ↦
+    H x (fun _ h ↦ IH _ (.of_mem_leftMoves h)) (fun _ h ↦ IH _ (.of_mem_rightMoves h))
+
+@[simp]
+theorem moveRecOn_eq {P : IGame → Sort*} (x)
+    (H : Π x, (Π y ∈ x.leftMoves, P y) → (Π y ∈ x.rightMoves, P y) → P x) :
+    moveRecOn x H = H x (fun y _ ↦ moveRecOn y H) (fun y _ ↦ moveRecOn y H) :=
+  isOption_wf.fix_eq ..
+
+-- TODO: docstring
```
tristan-f-r: This should be 'proper subposition': a 'subposition' would be defined over `Relation.ReflTransGen IsOption`.
I'm unaware of any use of Subsequent.

CombinatorialGames/IGame/IGame.lean
```diff
@@ -0,0 +1,529 @@
...
+
+@[cases_eliminator]
+theorem ind {P : IGame → Prop} (H : ∀ y, P (mk y)) (x : IGame) : P x :=
+  Quotient.ind H x
+
+/-- Choose an element of the equivalence class using the axiom of choice. -/
+def out (x : IGame) : PGame := Quotient.out x
+@[simp] theorem out_eq (x : IGame) : mk x.out = x := Quotient.out_eq x
+
+-- TODO: docstring
```
tristan-f-r:
```suggestion
/--
`Quotient.lift` for `IGame`: if f : PGame → α respects the identical equivalence relation ≡, then f
lifts IGame to α
-/
```

LeoDog896 (changes requested):
The proofs here are incredible! Just some docstring suggestions and a note on the name of `Subsequent`.

CombinatorialGames/IGame/IGame.lean
vihdzp: Making this reflexive sort of defeats its purpose, since it becomes no longer well-founded. Otherwise I agree with renaming this to `Subposition`.

CombinatorialGames/IGame/IGame.lean
vihdzp: On second consideration, I only use this definition twice, in order to define `IGame.leftMoves` and `IGame.rightMoves`. Since we want to expose `PGame` as little as possible within the API, I've simply inlined `Quotient.lift` into their definitions.

LeoDog896 (approved):

PR #43, opened 2025-03-02
feat: `Ordinal.toIGame`

vihdzp:
I modified the API a bit, making `Ordinal.toIGame` an `OrderEmbedding`, in exchange for having to prove a few extra auxiliary theorems.

### DIFF
--- a/CombinatorialGames.lean
+++ b/CombinatorialGames.lean
@@ -1,16 +1,3 @@
-import CombinatorialGames.Counterexamples.Multiplication
-import CombinatorialGames.Game.Basic
-import CombinatorialGames.Game.Birthday
-import CombinatorialGames.Game.Concrete
-import CombinatorialGames.Game.Impartial
-import CombinatorialGames.Game.Ordinal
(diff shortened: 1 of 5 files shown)


vihdzp:
CI fails due to the duplicated `Ordinal.toGame`. The thing I had been doing previously is putting everything in a temp namespace. But that doesn't work here since `Ordinal` is defined (in the root namespace) in Mathlib.

What should we do?

LeoDog896:
> What should we do? A (somewhat cheeky) idea I have is to change the root file to import only the refactored files. That is, we cut off all of the prior `PGame` API and gradually add it back in terms of `IGame` as we refactor files. (In particular, we can get rid of the temp namespaces).

That sounds good to me - cutting off the legacy `PGame` files is perfectly reasonable for this refactor :+1:

vihdzp:
Almost forgot about doing this, but I'll implement my plan of making the codomain `NatOrdinal` rather than `Ordinal`.

LeoDog896 (approved):

PR #51, opened 2025-03-05
feat: `Short x` → `Short y` → `Short (x * y)`

vihdzp:
The games generated are so big it's basically impossible to make use of them. Oh well.


CombinatorialGames/IGame/Short.lean
```diff
@@ -75,6 +75,26 @@ macro "sgame_wf" : tactic =>
     [Prod.Lex.left, Prod.Lex.right, PSigma.Lex.left, PSigma.Lex.right,
     IsOption.moveLeft, IsOption.moveRight] )
 
+def toStringAux {n : ℕ} (f : Fin n → String) : String :=
```
tristan-f-r: Despite this being an auxiliary definition, I am uncomfortable with the name being `toStringAux` despite it acting on `Fin n -> String` instead of `SGame`.

CombinatorialGames/IGame/Short.lean
```diff
@@ -75,6 +75,26 @@ macro "sgame_wf" : tactic =>
...
+def toString' (x : SGame) : String :=
+  "{" ++ toStringAux (fun i ↦ toString' (x.moveLeft i)) ++ " | " ++
+    toStringAux (fun j ↦ toString' (x.moveRight j)) ++ "}"
+termination_by x
+decreasing_by sgame_wf
+
+/-- The size of an `SGame` is its number of proper subpositions.
+
+Note that this doesn't necessarily correspond to the number of subpositions of the `IGame`, though
+it does serve as a way to estimate how "complex" the game is to build. -/
```
tristan-f-r:
```suggestion
it does serve as a way to estimate how "complex" the game was to build. -/
```

CombinatorialGames/IGame/Short.lean
```diff
@@ -333,7 +354,130 @@ theorem toIGame_add (x y : SGame) : toIGame (x + y) = toIGame x + toIGame y := b
...
+theorem toIGame_mul (x y : SGame) : toIGame (x * y) = toIGame x * toIGame y := by
+  ext z
+  on_goal 1 =>
+    simp only [leftMoves_toIGame, Set.mem_range]
+    let e : Fin (x * y).leftMoves ≃
+      Fin x.leftMoves × Fin y.leftMoves ⊕ Fin x.rightMoves × Fin y.rightMoves :=
+    (Equiv.cast (by simp)).trans <|
+      finSumFinEquiv.symm.trans (Equiv.sumCongr finProdFinEquiv.symm finProdFinEquiv.symm)
+    rw [e.exists_congr_left]
+    simp [e, -finProdFinEquiv_symm_apply]
```
tristan-f-r, line 455: There's a lot of non-terminal simps here - is it safe to use them before `trans`?

LeoDog896 (changes requested):

CombinatorialGames/IGame/Short.lean
vihdzp: I'll simply inline this into `toString'`.

CombinatorialGames/IGame/Short.lean
vihdzp: In theory probably not? But if the `simp` ever breaks (which I find unlikely, since it only uses relatively basic lemmas), all we need to do is rewrite the expression in the `trans` block to match. 

Writing down the `simp only` in full takes four entire lines of code. Very nasty proposition.

vihdzp: FYI what's happening here is the following. Calling `simp` leaves an expression that looks like `∃ a b, a = f x ∧ ...`. Unfortunately, `simp` isn't able to apply the nested substitution. So instead we write down what the goal would look like if we did. We prove that the goal is equivalent through the more powerful `aesop`. And then we finally conclude by congruence + the inductive hypothesis.

PR #58, opened 2025-03-06
feat: `game_cmp` tactic

vihdzp:
We implement a simple tactic capable of proving simple inequalities on games. The idea is to use this for computation, instead of trying to square the circle through barely functional decidability tactics.


LeoDog896 (approved):
(There is a merge conflict present.)

LeoDog896 (approved):

PR #59, opened 2025-03-06
feat: redefine `Short`

vihdzp:
After some discussion, it seems like the approach of trying to shoehorn decidability into `Short` isn't the best. We redefine it in the more typical mathematical manner, and we do some API battle testing by proving that basic arithmetic preserves shortness, and that short games are bounded by some natural number.

- [x] depends on: #58


CombinatorialGames/IGame/Short.lean
```diff
@@ -4,560 +4,171 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
-it does serve as a way to estimate how "complex" the game was to build. -/
-def size (x : SGame) : ℕ :=
-  ∑ i, ((x.moveLeft i).size + 1) + ∑ j, ((x.moveRight j).size + 1)
-termination_by x
-decreasing_by sgame_wf
-
-instance : ToString SGame := ⟨toString'⟩
-
-/-! ### Decidability instances -/
+namespace IGame
```
vihdzp, line 21: This diff is really ugly, I'd recommend you just read the new file on its own.

CombinatorialGames/IGame/Short.lean
```diff
@@ -4,560 +4,175 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
-Note that this doesn't necessarily correspond to the number of subpositions of the `IGame`, though
-it does serve as a way to estimate how "complex" the game was to build. -/
-def size (x : SGame) : ℕ :=
-  ∑ i, ((x.moveLeft i).size + 1) + ∑ j, ((x.moveRight j).size + 1)
+theorem finite_setOf_subposition (x : IGame) [Short x] : {y | Subposition y x}.Finite := by
+  have : {y | Subposition y x} = {y | IsOption y x} ∪
+      ⋃ y ∈ {y | IsOption y x}, {z | Subposition z y} := by
+    ext
+    rw [Set.mem_setOf_eq, Subposition, Relation.transGen_iff]
+    simp [and_comm]
```
tristan-f-r, line 80: (This feels like a good use case for suffices, but I couldn't get that to work.)

LeoDog896 (approved):

PR #63, opened 2025-03-08
feat: poset `IGame`

LeoDog896:
Note: `rel_leftMove_iff` and associated lemmas were made because we can't benefit from `toIGame` being isomorphic.


vihdzp:
If we're not using the "subsequent" terminology we should probably rename `subsequentL` and `subsequentR`. What about `relLeft` and `relRight`?

CombinatorialGames/IGame/Concrete.lean
```diff
@@ -71,6 +72,20 @@ theorem leftMoves_toIGame (a : α) : (toIGame a).leftMoves = toIGame '' {b | b 
 theorem rightMoves_toIGame (a : α) : (toIGame a).rightMoves = toIGame '' {b | b ≺ᵣ a} := by
   rw [toIGame_def, rightMoves_ofSets]
 
+theorem rel_leftMove_iff {a b : α} :
+    toIGame b ∈ (toIGame a).leftMoves ↔ ∃ c, c ≺ₗ a ∧ toIGame c = toIGame b := by
+  simp_rw [leftMoves_toIGame, Set.mem_image, Set.mem_setOf_eq]
```
vihdzp: I don't think we need this lemma, since this works:
```suggestion
theorem rel_leftMove_iff {a b : α} :
    toIGame b ∈ (toIGame a).leftMoves ↔ ∃ c, c ≺ₗ a ∧ toIGame c = toIGame b := by
  rw [leftMoves_toIGame]; rfl
```

CombinatorialGames/IGame/Concrete.lean
```diff
@@ -71,6 +72,20 @@ theorem leftMoves_toIGame (a : α) : (toIGame a).leftMoves = toIGame '' {b | b 
...
+
+theorem rel_leftMove {a b : α} (hab : b ≺ₗ a) : toIGame b ∈ (toIGame a).leftMoves :=
```
vihdzp:
```suggestion
theorem mem_leftMoves_toIGame_of_relLeft {a b : α} (hab : b ≺ₗ a) : toIGame b ∈ (toIGame a).leftMoves :=
```
Somewhat wordy name but it matches e.g. `NatOrdinal.mem_leftMoves_toIGame_of_lt`.

vihdzp (commented):
Otherwise LGTM.

vihdzp:
We should change the field names in the `ConcreteGame` structure as well.

LeoDog896:
It should be moved back to `subsequentL/R` since I believe that's more descriptive than `relL` and `relR`.

vihdzp (approved):

PR #65, opened 2025-03-09
feat: port mult counterexample

LeoDog896:
thanks to `game_cmp`, the proofs here have been substantially shortened!

### DIFF
--- a/CombinatorialGames/Counterexamples/Multiplication.lean
+++ b/CombinatorialGames/Counterexamples/Multiplication.lean
@@ -46,6 +46,6 @@ theorem star'_mul_star_lf : star' * star ⧏ star := by
 
 /-- Pre-game multiplication cannot be lifted to games. -/
 theorem mul_not_lift : ∃ x₁ x₂ y : IGame, x₁ ≈ x₂ ∧ ¬ x₁ * y ≈ x₂ * y :=
-  ⟨star', ⋆, ⋆, star'_equiv_star, fun h ↦ absurd (h.trans star_sq).ge star'_mul_star_lf⟩
+  ⟨_, _, _, star'_equiv_star, fun h ↦ absurd (h.trans star_sq).ge star'_mul_star_lf⟩
 
 end IGame
--- a/CombinatorialGames/Counterexamples/Multiplication.lean
+++ b/CombinatorialGames/Counterexamples/Multiplication.lean
@@ -3,8 +3,8 @@ Copyright (c) 2024 Violeta Hernández Palacios. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Violeta Hernández Palacios
 -/
-import CombinatorialGames.Game.Basic
-import CombinatorialGames.Game.Special
+import CombinatorialGames.IGame.Basic
+import CombinatorialGames.IGame.Special
 import Mathlib.Tactic.FinCases
 
 /-!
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Counterexamples/Multiplication.lean
```diff
@@ -18,61 +18,34 @@ is colloquially known as `star`, so we use the name `star'` for the second. We p
...
-  case' right =>
-    dsimp
-    rw [← neg_le_neg_iff]
-    apply (neg_mul _ _).symm.equiv.1.trans
-    rw [neg_star]
-  assumption'
+theorem star_sq : ⋆ * ⋆ ≈ ⋆ := by game_cmp
 
 /-- `*'* ⧏ *` implies `*'* ≉ *`.-/
 theorem star'_mul_star_lf : star' * star ⧏ star := by
```
vihdzp, line 42: It's actually possible to solve this using the `game_cmp` tactic... after tweaking some `simp` lemmas:
```lean 4
@[simp]
theorem exists_imp {α} {P Q : α → Prop} :
    (∃ x, P x → Q x) ↔ (∃ x, ¬ P x) ∨ (∃ x, Q x) := by
  simp_rw [imp_iff_not_or, exists_or]

attribute [simp] exists_or

/-- `*'* ⧏ *` implies `*'* ≉ *`.-/
theorem star'_mul_star_lf : star' * star ⧏ star := by game_cmp
```
I think it's worth exploring what can be done to make this tactic more powerful. Currently, its main limitation is that it gets stuck on goals of the form `∃ x, ... (x = y) ...` or `∀ x, ... (x = y) ...`, which quantify over a variable which we actually know the value of due to a nested equality. Perhaps the way to remedy this is to greedily try and "split" quantifiers no matter what? Turning for instance `∃ x, P x ∨ Q x` into `(∃ x, P x) ∨ (∃ x, Q x)` until hopefully, all the predicates in quantifiers are equalities which `simp` can just get rid of? Or maybe the actual solution is to just wait for someone to create a simpproc for these sorts of nested equalities.

I think my solution is somewhat hacky, and your solution is just three lines of code anyways, so let's just merge this and think about this separately.

vihdzp (approved):

CombinatorialGames/Counterexamples/Multiplication.lean
```diff
@@ -18,61 +18,34 @@ is colloquially known as `star`, so we use the name `star'` for the second. We p
...
-      fin_cases i
-      · exact zero_lf_star
-      · exact (neg_lt_zero_iff.2 PGame.zero_lt_one).trans_lf zero_lf_star
-    · exact fun _ ↦ lf_zero_le.2 ⟨⟨0, Nat.zero_lt_two⟩, le_rfl⟩
-  constructor
-  case' right => dsimp; rw [← neg_le_neg_iff, neg_star, neg_star']
-  assumption'
+theorem star'_equiv_star : star' ≈ ⋆ := by game_cmp
 
 /-- The equation `** = *` is an identity, though not a relabelling. -/
```
vihdzp, line 38: Whoops, forgot to mention that "relabelling" isn't really a concept we're using anymore. In fact, even though it's not really relevant to the counterexample, it might be worth proving that `** = *` equality.

CombinatorialGames/Counterexamples/Multiplication.lean
tristan-f-r: By the way, what you suggested is exactly how I tried to do it when proving this: I was surprised that `aesop` didn't do that automatically, but I did just permute every one of the choices since there were finitely many, and it was quicker than loading up the previous Counterexamples file and trying to find the used numbers.

Needless to say, that's a good idea for the tactic :+1:

PR #67, opened 2025-03-10
chore: drop old Game folder

LeoDog896:
now that `Poset` is ported, we have less baggage: since `Birthday` was already deleted in its porting PR, we can drop everything now (I suppose that porting would require reimporting this, which hopefully isn't too much of a hassle.)


LeoDog896:
Since the actual PR diff is horrible, its best to look at each individual commit.

vihdzp:
Fyi, there's a merge conflict here.

vihdzp:
I've deleted the dyadic file. We've been discussing in DMs about how this file should be reworked, and I'm willing to take on that job in the next week or so. I'll also write down an issue about this so that this discussion doesn't just stay in private.

vihdzp:
I *think* I didn't break anything, so I'll just merge as is, and if something did go wrong we can fix it in a subsequent PR.

PR #71, opened 2025-03-11
feat: better examples of combinatorial games

vihdzp:
As pointed out on [Zulip](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Combinatorial.20game.20theory.20repository/near/504973135), tic-tac-toe and chess aren't actually combinatorial games in the sense this repository is interested in, as they can end in ties.

(Are there better-known games we can use as examples? Chomp and Hackenbush seem to mostly only be played within mathematical circles)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,9 @@ A formalization of topics within combinatorial game theory in Lean 4.
 
 ## What is it?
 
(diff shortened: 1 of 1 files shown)


LeoDog896:
Sim, Toads and Frogs?

vihdzp:
I'm not sure if I'd classify those as games played outside of mathematical spheres. If we're just trying to list important examples, I think the current list is probably fine.

LeoDog896 (approved):

PR #74, opened 2025-03-14
feat: surreal number from left and right sets

vihdzp:
I'm a bit unsure as to what API we should have for `Surreal.ofSets`, but I think that should become clearer once we start making use of it (this is needed e.g. to define `ω ^ x`).

### DIFF
--- a/CombinatorialGames/Game/Basic.lean
+++ b/CombinatorialGames/Game/Basic.lean
@@ -61,8 +61,8 @@ theorem equiv_mk_out (x : IGame) : x ≈ (mk x).out := (mk_out_equiv x).symm
 This is given notation `{s | t}ᴳ`, where the superscript `G` is to disambiguate from set builder
 notation, and from the analogous constructor on `IGame`.
 
-Note that although this function is well-defined, recovering the left/right sets from a game is not,
-as there are many sets that can generate a single game. -/
+Note that although this function is well-defined, this function isn't injective, nor do equivalence
+classes in `Game` have a canonical representative.  -/
 def ofSets (s t : Set Game.{u}) [Small.{u} s] [Small.{u} t] : Game.{u} :=
   mk {out '' s | out '' t}ᴵ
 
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Surreal/Basic.lean
```diff
@@ -371,5 +371,36 @@ theorem toGame_sub (x y : Surreal) : toGame (x - y) = toGame x - toGame y :=
...
+@[simp]
+theorem game_out_eq (x : Surreal) : Game.mk x.out = x.toGame := by
+  cases x
+  rw [toGame_mk, Game.mk_eq_mk]
+  exact mk_out_equiv _
+
+/-- Construct a `Surreal` from its left and right sets, and a proof that all elements from the left
+set are less than all the elements of the right set.
+
+Note that although this function is well-defined, recovering the left/right sets from a surreal
```
tristan-f-r: Perhaps be more clear about the lack of Surreal representatives for Surreals of infinite birthday?

... is well-defined, this function isn't injective nor do equivalence classes in `Surreal` have a well-defined representative ...

CombinatorialGames/Game/Basic.lean
```diff
@@ -68,6 +68,7 @@ def ofSets (s t : Set Game.{u}) [Small.{u} s] [Small.{u} t] : Game.{u} :=
 
 @[inherit_doc] notation "{" s " | " t "}ᴳ" => ofSets s t
 
+@[simp]
```
tristan-f-r, line 71: I'm uncomftorable with this being a `simp` tagged theorem, though this could be because I have a limited understanding of when this applies.

Would this apply more eagerly than congruence where `mk {s | t}ᴵ = mk {l | r}ᴵ iff {s | t}ᴵ = {l | r}ᴵ`?

LeoDog896 (changes requested):

CombinatorialGames/Surreal/Basic.lean
vihdzp: I've updated the docstring.

CombinatorialGames/Game/Basic.lean
vihdzp: I think it's somewhat unlikely that we'd have a `simp` lemma of the form `{s | t}ᴵ = {l | r}ᴵ`. In the case where `s = l` and `t = r`, the presence of this lemma shouldn't prevent `simp` from closing the goal.

LeoDog896 (approved):

PR #82, opened 2025-03-18
feat: options of rational are also rational

vihdzp:
One could argue that this exposes internal details about `RatCast`, but in my view, any reasonable definition of a rational number as a game should satisfy this property. Besides, this is useful in proving that e.g. rational numbers are equivalent to their Dedekind cuts of dyadic rationals.
- [x] depends on: #81
- [x] depends on: #83


CombinatorialGames/Surreal/Division.lean
```diff
@@ -301,3 +320,33 @@ theorem mk_ratCast (q : ℚ) : mk q = q := by
...
+namespace IGame
+
+-- TODO: upstream
+attribute [simp] AntisymmRel.refl
+
+private theorem equiv_ratCast_of_mem_move_inv_natCast {n : ℕ} :
+    (∀ x ∈ leftMoves.{u} (n + 1)⁻¹, ∃ q : ℚ, x ≈ q) ∧
+      (∀ x ∈ rightMoves.{u} (n + 1)⁻¹, ∃ q : ℚ, x ≈ q) := by
+  refine invRec _ ⟨0, ?_⟩ ?_ ?_ ?_ ?_
+  any_goals simp
```
vihdzp: This solves three of the five goals, and is a non-terminal simp in the other two, which happens to simplify favorably. I'm not aware of any way to write down "`simp`, and if it doesn't solve the goal, fail", but this should probably be replaced by that.

tristan-f-r: You could use `try · simp`. (Made a suggestion below)

CombinatorialGames/Surreal/Division.lean
```diff
@@ -291,4 +321,56 @@ theorem ratCast_strictMono : StrictMono ((↑) : ℚ → IGame) :=
...
+  cases n with
+  | zero => simp
+  | succ n =>
+    refine invRec (mod_cast n.succ_pos) ⟨0, ?_⟩ ?_ ?_ ?_ ?_
+    any_goals simp
+    all_goals
+      rintro _ hn rfl x hx q hq
+      use (1 + -q) / n
+      first | have := Numeric.of_mem_leftMoves hx | have := Numeric.of_mem_rightMoves hx
+      simp_all [invOption, ← Surreal.mk_eq_mk]
```
tristan-f-r:
```suggestion
private theorem equiv_ratCast_of_mem_move_inv_natCast {n : ℕ} :
    (∀ x ∈ leftMoves.{u} n⁻¹, ∃ q : ℚ, x ≈ q) ∧ (∀ x ∈ rightMoves.{u} n⁻¹, ∃ q : ℚ, x ≈ q) := by
  cases n with
  | zero => simp
  | succ n =>
    refine invRec (mod_cast n.succ_pos) ⟨0, ?_⟩ ?_ ?_ ?_ ?_
    any_goals (try · simp)
    all_goals
      simp_rw [Nat.cast_add, Nat.cast_one, leftMoves_natCast_succ, forall_exists_index]
      rintro _ hn rfl x hx q hq
      use (1 + -q) / n
      first | have := Numeric.of_mem_leftMoves hx | have := Numeric.of_mem_rightMoves hx
      simp_all [invOption, ← Surreal.mk_eq_mk]
```

LeoDog896 (changes requested):

CombinatorialGames/Surreal/Division.lean
vihdzp: Wait, can't you just do `any_goals · simp`?

tristan-f-r: That should work too!

vihdzp: Doesn't work for whatever reason. I replaced this with `<;> (try · simp)`, since `any_goals` + `try` is redundant.

LeoDog896 (approved):

PR #83, opened 2025-03-18
feat: better junk values for `IGame` inverse

vihdzp:
I originally intended to define `x⁻¹` on `IGame` only for positive games, leaving the inverse on negative games as just some junk value. This was a deliberate choice, since it reflects that the genetic definition for the inverse only works for positive games. However, this introduces persistent assumptions `0 < x` all over the place.

We instead redefine `x⁻¹` so that it satisfies `0⁻¹ = 0` and `(-x)⁻¹ = -x⁻¹`. This introduces extra `0 < x` assumptions on the most basic theorems for `x⁻¹`, but allows us to get rid of this assumption for everything afterwards. In particular, we now have `Numeric x → Numeric x⁻¹` and `Numeric x → ¬ x ≈ 0 → x * x⁻¹ ≈ 1`.

### DIFF
--- a/CombinatorialGames/Game/IGame.lean
+++ b/CombinatorialGames/Game/IGame.lean
@@ -1229,7 +1229,9 @@ private theorem inv_eq {x : IGame.{u}} (hx : 0 < x) :
 
 protected theorem div_eq_mul_inv (x y : IGame) : x / y = x * y⁻¹ := rfl
 
-@[simp] protected theorem inv_zero : (0 : IGame)⁻¹ = 0 := by simp [inv_eq']
+theorem inv_of_equiv_zero {x : IGame} (h : x ≈ 0) : x⁻¹ = 0 := by simp [inv_eq', h.not_lt, h.not_gt]
+
+@[simp] protected theorem inv_zero : (0 : IGame)⁻¹ = 0 := inv_of_equiv_zero .rfl
 @[simp] protected theorem zero_div (x : IGame) : 0 / x = 0 := zero_mul _
 @[simp] protected theorem neg_div (x y : IGame) : -x / y = -(x / y) := neg_mul ..
 
@@ -1342,14 +1344,11 @@ theorem invRec {x : IGame} (hx : 0 < x)
       Q _ (invOption_right_right_mem_rightMoves_inv hx hy hyx ha)) :
     (∀ y (hy : y ∈ x⁻¹.leftMoves), P y hy) ∧ (∀ y (hy : y ∈ x⁻¹.rightMoves), Q y hy) := by
   apply invRec' hx zero
-  · convert left₁ using 6 with _ ha
-    simp_rw [invOption_eq ha]
-  · convert left₂ using 6 with _ ha
-    simp_rw [invOption_eq ha]
-  · convert right₁ using 6 with _ ha
-    simp_rw [invOption_eq ha]
-  · convert right₂ using 6 with _ ha
-    simp_rw [invOption_eq ha]
+  on_goal 1 => convert left₁ using 6 with _ ha
+  on_goal 2 => convert left₂ using 6 with _ ha
+  on_goal 3 => convert right₁ using 6 with _ ha
+  on_goal 4 => convert right₂ using 6 with _ ha
+  all_goals simp_rw [invOption_eq ha]
 
 instance : RatCast IGame where
   ratCast q := q.num / q.den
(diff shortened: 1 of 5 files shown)


CombinatorialGames/Game/IGame.lean
```diff
@@ -1188,128 +1190,165 @@ private def InvTy.val' {x : IGame}
...
+  rw [inv_eq', inv_eq']
+  obtain h | h | h | h := lt_or_antisymmRel_or_gt_or_incompRel x 0
+  repeat
+    simp [h, h.asymm]
+    simp [h.not_lt, h.not_gt]
 
 /-- The general option of `x⁻¹` looks like `(1 + (y - x) * a) / y`, for `y` an option of `x`, and
 `a` some other "earlier" option of `x⁻¹`. -/
 @[pp_nodot]
 def invOption (x y a : IGame) : IGame :=
```
vihdzp, line 1249: An annoying consequence of this refactor is that the `InvTy.val` values are no longer def-eq to `invOption`, which is why this section becomes longer. Of course, this doesn't really matter once we prove `invRec`.

CombinatorialGames/Surreal/Division.lean
```diff
@@ -210,94 +220,55 @@ end Surreal.Division
...
-  have := Numeric.inv hy'
-  have := (Numeric.mul_inv_cancel hx).trans (Numeric.mul_inv_cancel hy').symm
-  rw [← (Numeric.mul_congr_left hy).antisymmRel_congr_right] at this
-  exact Numeric.mul_left_cancel hx.not_antisymmRel_symm this
-
-protected instance ratCast (q : ℚ) : Numeric q := by
-  have : Numeric (q.den : IGame)⁻¹ := .inv (mod_cast q.den_pos)
-  change Numeric (_ * _)
-  infer_instance
+protected instance inv {x : IGame} [Numeric x] : Numeric x⁻¹ := by
```
vihdzp: Here's where the magic happens!

CombinatorialGames/Surreal/Division.lean
```diff
@@ -146,42 +156,42 @@ lemma option_mul_inv_lt {x : IGame} [Numeric x] (hx : 0 < x)
...
     (∀ y ∈ (x * x⁻¹).leftMoves, y < 1) ∧ (∀ y ∈ (x * x⁻¹).rightMoves, 1 < y) := by
   have := numeric_inv hx hl hr hl' hr'
-  obtain ⟨Hl, Hr⟩ := numeric_option_inv hl hr
+  obtain ⟨Hl, Hr⟩ := numeric_option_inv hx hl hr
   rw [forall_leftMoves_mul, forall_rightMoves_mul]
   refine ⟨⟨?_, ?_⟩, ⟨?_, ?_⟩⟩
   all_goals
-    intro y hy a ha
-    first | have := Numeric.of_mem_leftMoves hy | have := Numeric.of_mem_rightMoves hy
+    intro y hyx a ha
```
vihdzp, line 163: This proof didn't really change that much, I just tried to enforce the naming convention that `hyx : y ∈ x.(left|right)Moves` and `hy : 0 < y` (since it's a bit weird to have `hx : 0 < x` and `hy' : 0 < y` otherwise)

CombinatorialGames/Game/IGame.lean
```diff
@@ -1303,13 +1322,33 @@ theorem invRec (x : IGame) {P : ∀ y ∈ x⁻¹.leftMoves, Prop} {Q : ∀ y ∈
...
+      Q _ (invOption_left_left_mem_rightMoves_inv hx hy hyx ha))
+    (right₂ : ∀ y (hy : 0 < y) (hyx : y ∈ x.rightMoves), ∀ a (ha : a ∈ x⁻¹.rightMoves), Q a ha →
+      Q _ (invOption_right_right_mem_rightMoves_inv hx hy hyx ha)) :
+    (∀ y (hy : y ∈ x⁻¹.leftMoves), P y hy) ∧ (∀ y (hy : y ∈ x⁻¹.rightMoves), Q y hy) := by
+  apply invRec' hx zero
+  on_goal 1 => convert left₁ using 6 with _ ha
+  on_goal 2 => convert left₂ using 6 with _ ha
+  on_goal 3 => convert right₁ using 6 with _ ha
+  on_goal 4 => convert right₂ using 6 with _ ha
+  all_goals simp_rw [invOption_eq ha]
```
tristan-f-r, line 1351: (Not any review comment, and if it were it would just be a nitpick [though I'm tempted to prefer `simp_rw` over `simp only` for this scenario] but I must ask: is there any meaningful difference between `simp_rw [h1]` and `simp only [h1]` past `simp_rw`'s ease of extension?)

vihdzp: I think they're the same, `simp_rw [h₁, h₂, h₃]` basically amounts to `simp only [h₁]; simp only [h₂]; simp only [h₃]`.

tristan-f-r (approved):

PR #92, opened 2025-03-25
feat: string nim

tristan-f-r:
we formalize the basics of https://doi.org/10.48550/arXiv.2503.16781 without additional theory

### DIFF
--- a/CombinatorialGames/Game/Specific/StrNim.lean
+++ b/CombinatorialGames/Game/Specific/StrNim.lean
@@ -33,37 +33,40 @@ instance : DecidableEq (StrNim σ) := instDecidableEqList -- deriving DecidableE
 
 namespace StrNim
 
+def length (s : StrNim σ) : ℕ := (ofStrNim s).length
+
 instance : EmptyCollection (StrNim σ) := ⟨toStrNim []⟩
 
 variable {σ : Type u} [DecidableEq σ] (s t : StrNim σ)
 
-abbrev char (i : ℕ) : Option σ := (ofStrNim s)[i]?
+instance : GetElem (StrNim σ) Nat σ fun s i => i < s.length where
+  getElem s i hi := (ofStrNim s)[i]'hi
 
 structure Move where
   idx : ℕ
   len : ℕ
   len_pos : 0 < len
-  bounded : idx + len < (ofStrNim s).length
+  bounded : idx + len < s.length
   equiv : ∃ σ, List.take len (List.drop idx (ofStrNim s)) = List.replicate len σ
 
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Game/Specific/StrNim.lean
```diff
@@ -0,0 +1,124 @@
...
+@[simp] theorem ofStrNim_toStrNim (l : List σ) :
+  ofStrNim (toStrNim l) = l := rfl
+
+namespace StrNim
+
+instance : EmptyCollection (StrNim σ) := ⟨toStrNim []⟩
+
+variable {σ : Type u} [DecidableEq σ] (s t : StrNim σ)
+
+abbrev char (i : ℕ) : Option σ := (ofStrNim s)[i]?
```
vihdzp: A better option might be to derive `GetElem` to enable the `s[i]` notation directly.

CombinatorialGames/Game/Specific/StrNim.lean
```diff
@@ -0,0 +1,124 @@
...
+
+structure Move where
+  idx : ℕ
+  len : ℕ
+  len_pos : 0 < len
+  bounded : idx + len < (ofStrNim s).length
```
vihdzp: Might be useful turning `(ofStrNim s).length` into its own auxiliary `length` function?

CombinatorialGames/Game/Specific/StrNim.lean
```diff
@@ -0,0 +1,124 @@
...
+  equiv : ∃ σ, List.take len (List.drop idx (ofStrNim s)) = List.replicate len σ
```
vihdzp, line 50: I suspect it might be nicer to write this down in terms of indices. That is, something like "for every `i` between the start and end, `s[i] = σ`". This setup would require that you use `start < end` as parameters for the move rather than `idx` and `len`, though.

CombinatorialGames/Game/Specific/StrNim.lean
```diff
@@ -0,0 +1,124 @@
...
+
+theorem Move.idx_bounded (m : Move s) : m.idx < (ofStrNim s).length - 1 := by
```
vihdzp: You can prove this stronger result:
```suggestion
theorem Move.idx_bounded (m : Move s) : m.idx < (ofStrNim s).length - m.len := by
```

CombinatorialGames/Game/Specific/StrNim.lean
tristan-f-r: It would be nicer definitionally wise to remove the use of `take` and `drop`, but I did it this way to reduce the number of hypotheses in the definition.

PR #93, opened 2025-04-14
ci: pages

tristan-f-r:
adds `docbuild` so we can have generated documentation on GitHub Pages! this may require configuring `Settings -> Pages` to allow actions, and giving more permissions to Actions (also in Settings).


docbuild/lean-toolchain
```diff
@@ -0,0 +1 @@
+leanprover/lean4:v4.18.0-rc1
```
vihdzp, line 1: Sorry for taking so long! You might want to update this value.

vihdzp (approved):
I don't really know much about CI, so I'm delegating this to you.

PR #94, opened 2025-06-21
feat: short games have ℕ birthdays

tristan-f-r:
This is proved via explicit (noncomputable) Finset construction, which can be later used to formalize some known values to https://oeis.org/A065401.

PR #95, opened 2025-06-29
feat: more theorems on `Short`

vihdzp:
These are useful for #94.

### DIFF
--- a/CombinatorialGames/Game/Short.lean
+++ b/CombinatorialGames/Game/Short.lean
@@ -59,6 +59,9 @@ instance (x : IGame) [Short x] : Finite x.leftMoves :=
 instance (x : IGame) [Short x] : Finite x.rightMoves :=
   (Short.finite_rightMoves x).to_subtype
 
+instance (x : IGame) [Short x] : Finite {y // IsOption y x} :=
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Game/Short.lean
```diff
@@ -72,6 +75,15 @@ protected theorem isOption [Short x] (h : IsOption y x) : Short y := by
 
 alias _root_.IGame.IsOption.short := Short.isOption
 
+protected theorem subposition {x : IGame} [Short x] (h : Subposition y x) : Short y := by
```
tristan-f-r, line 78:
```suggestion
protected theorem subposition {x y : IGame} [Short x] (h : Subposition y x) : Short y := by
```

tristan-f-r (approved):
These theorems look good :+1:. Just one comment, but I'll delegate this to you.

CombinatorialGames/Game/Short.lean
vihdzp: `y` already exists as an implicit variable. So does `x`, but we need to specify it explicitly so that it's generalized in our induction.

PR #98, opened 2025-06-29
feat: define `Dyadic.toIGame`

vihdzp:
We define the canonical map from the dyadic numbers into `IGame` and prove some basic properties.

This PR really is more about the dyadic rational numbers than about game theory. Since Mathlib doesn't have the dyadic rationals, we have to set up a lot of stuff on our own. Note the conscious decision not to use `Localization.Away 2` - at this point, using it would be much more trouble than it'd be worth.

### DIFF
--- a/CombinatorialGames/Surreal/Numeric.lean
+++ b/CombinatorialGames/Surreal/Numeric.lean
@@ -258,11 +258,9 @@ theorem toIGame_neg (x : Dyadic) : toIGame (-x) = -toIGame x := by
   dsimp
   split_ifs with h
   · simp
-  · simp only [lower_neg, upper_neg, IGame.neg_ofSets, Set.neg_singleton, IGame.ofSets_inj,
-      Set.singleton_eq_singleton_iff]
-    have := den_upper_lt h
+  · have := den_upper_lt h
     have := den_lower_lt h
-    exact ⟨toIGame_neg _, toIGame_neg _⟩
+    simpa using ⟨toIGame_neg _, toIGame_neg _⟩
 termination_by x.1.den
 
 end Dyadic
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -0,0 +1,268 @@
...
+theorem toIGame_neg (x : Dyadic) : toIGame (-x) = -toIGame x := by
+  unfold toIGame
+  dsimp
+  split_ifs with h
+  · simp
+  · simp only [lower_neg, upper_neg, IGame.neg_ofSets, Set.neg_singleton, IGame.ofSets_inj,
+      Set.singleton_eq_singleton_iff]
+    have := den_upper_lt h
+    have := den_lower_lt h
+    exact ⟨toIGame_neg _, toIGame_neg _⟩
```
tristan-f-r:
```suggestion
  · have := den_upper_lt h
    have := den_lower_lt h
    simpa using ⟨toIGame_neg _, toIGame_neg _⟩
```

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -0,0 +1,268 @@
...
+  obtain ⟨n, rfl⟩ := hy
+  have := Rat.den_dvd x (2 ^ n)
+  rw [dvd_prime_pow Int.prime_two] at this
+  obtain ⟨m, -, hm⟩ := this
+  obtain hm | hm := Int.associated_iff.1 hm
+  · use m
+    rw [Rat.mkRat_eq_divInt, Nat.cast_pow, Nat.cast_ofNat, ← Nat.cast_inj (R := ℤ), hm]
+    simp
+  · rw [← Nat.cast_ofNat, ← Nat.cast_pow 2 m, Nat.cast_eq_neg_cast] at hm
+    aesop
```
tristan-f-r:
```suggestion
    simp_all
```

CombinatorialGames/Surreal/Numeric.lean
tristan-f-r: (If `simpa` was avoided here for good reason, at the very least that long `simp only` call should be `simp_rw`)

PR #100, opened 2025-06-30
feat: `Numeric (Dyadic.toIGame x)`

vihdzp:
We prove that dyadic games are in fact numeric, and that they embed the dyadic rationals into the poset of games.

I came up with this induction argument myself. If you know of a simpler proof then lmk.

- [x] Depends on: #99


tristan-f-r (approved):
If there was an earlier construction for `Dyadic` as a ring homomorphism to `IGame`, then you could prove `toIGame_lt_toIGame` after showing `0 < x ↔ 0 < toIGame x`. Since that is not the case, though, this alternative proof works just as well :+1:

PR #102, opened 2025-06-30
feat: delabs for equiv and fuzzy rel

plp127:


vihdzp (approved):
I don't know much about delaborators, but this seems correct to me, and it seems to work.

PR #107, opened 2025-07-02
feat(Dyadic.toIGame): preserves addition, order embedding, and numeric range

plp127:
Prove a 3-in-1 theorem
- `Numeric (toIGame x)`
- `toIGame x ≤ toIGame y ↔ x ≤ y`
- `toIGame (x + y) ≈ toIGame x + toIGame y`

### DIFF
--- a/CombinatorialGames/Surreal/Numeric.lean
+++ b/CombinatorialGames/Surreal/Numeric.lean
@@ -94,6 +94,22 @@ theorem IsDyadic.add {x y : ℚ} (hx : IsDyadic x) (hy : IsDyadic y) : IsDyadic 
   rw [Rat.add_def']
   exact IsDyadic.mkRat _ (Submonoid.mul_mem _ hx hy)
 
+theorem IsDyadic.nsmul {x : ℚ} (n : ℕ) (hx : IsDyadic x) : IsDyadic (n • x) := by
+  induction n with
+  | zero => exact ⟨0, rfl⟩
+  | succ n ih =>
+    rw [succ_nsmul]
+    exact ih.add hx
+
+theorem IsDyadic.zsmul {x : ℚ} (n : ℤ) (hx : IsDyadic x) : IsDyadic (n • x) := by
+  cases n with
+  | ofNat n =>
+    rw [Int.ofNat_eq_natCast, natCast_zsmul]
+    exact hx.nsmul n
+  | negSucc n =>
+    rw [negSucc_zsmul]
+    exact (hx.nsmul (n + 1)).neg
+
 theorem IsDyadic.sub {x y : ℚ} (hx : IsDyadic x) (hy : IsDyadic y) : IsDyadic (x - y) := by
   rw [sub_eq_add_neg]
   exact hx.add hy.neg
@@ -101,6 +117,13 @@ theorem IsDyadic.sub {x y : ℚ} (hx : IsDyadic x) (hy : IsDyadic y) : IsDyadic 
 theorem IsDyadic.mul {x y : ℚ} (hx : IsDyadic x) (hy : IsDyadic y) : IsDyadic (x * y) := by
   rw [Rat.mul_def, Rat.normalize_eq_mkRat]
   exact IsDyadic.mkRat _ (Submonoid.mul_mem _ hx hy)
+
+theorem IsDyadic.pow {x : ℚ} (hx : IsDyadic x) (n : ℕ) : IsDyadic (x ^ n) := by
+  induction n with
+  | zero => exact ⟨0, rfl⟩
+  | succ n ih =>
+    rw [pow_succ]
+    exact ih.mul hx
 
 /-- The subtype of `IsDyadic` numbers.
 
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -94,6 +94,22 @@ theorem IsDyadic.add {x y : ℚ} (hx : IsDyadic x) (hy : IsDyadic y) : IsDyadic
   rw [Rat.add_def']
   exact IsDyadic.mkRat _ (Submonoid.mul_mem _ hx hy)
 
+theorem IsDyadic.nsmul {x : ℚ} (n : ℕ) (hx : IsDyadic x) : IsDyadic (n • x) := by
```
vihdzp, line 97: Aren't this and `IsDyadic.zsmul` easier to prove through `IsDyadic.mul`?

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -120,7 +143,7 @@ instance : NatCast Dyadic where
 instance : IntCast Dyadic where
   intCast n := ⟨n, ⟨0, rfl⟩⟩
 
-@[simp] theorem val_intCast (n : ℕ) : (n : Dyadic).val = n := rfl
+@[simp] theorem val_intCast (n : ℤ) : (n : Dyadic).val = n := rfl
```
vihdzp, line 146: Embarrassing typo on my end, thanks!

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -193,6 +234,9 @@ theorem odd_num {x : Dyadic} (hx : x.1.den ≠ 1) : Odd x.1.num := by
   rw [← Int.natAbs_dvd_natAbs]
   exact (Nat.not_coprime_of_dvd_of_dvd one_lt_two · hd x.1.reduced)
 
+instance : CanLift Dyadic Int Int.cast (·.1.den = 1) where
```
vihdzp, line 237: I think everything up to here in the PR are good additions, regardless of whether the rest is redundant or not.

vihdzp (commented):
I already had a proof sketch of `toIGame (x + y) ≈ toIGame x + toIGame y` planned out. It uses the simplicity theorem, proving that `toIGame x + toIGame y` fits in `toIGame (x + y)`, but fits in none of its options. We induct in the lexicographically ordered pair `(toIGame x, toIGame y)`, which lets us assume that any option of `toIGame x + toIGame y` is equivalent to some `toIGame z`, at which point we can use the theorem on comparisons to convert the problem into basic fractions.

Note that once we have these three results, we're able to easily prove `toIGame x ≈ ↑x.val` by induction on the denominator, whence it follows that `fun x ↦ ⟦toIGame x⟧` is a ring homomorphism.

My PR exists in an incomplete form in the [`vi.numeric_pt3`](https://github.com/vihdzp/combinatorial-games/tree/vi.numeric_pt3) branch. I'll try to complete it today; if my proof sketch actually works, we can then figure out what the best parts of both PRs are.

CombinatorialGames/Surreal/Numeric.lean
plp127: Should I split those off into another PR?

CombinatorialGames/Surreal/Numeric.lean
plp127: This is how I usually prove stuff about nsmul. I though about going through `IsDyadic.mul` but I wasn't completely sure that it would work and I was completely sure that induction would work.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: Please do!

CombinatorialGames/Surreal/Numeric.lean
vihdzp: `n • x = ↑n * x`, and both of these are dyadic.

CombinatorialGames/Surreal/Numeric.lean
plp127: #108

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -226,6 +270,143 @@ theorem upper_neg (x : Dyadic) : upper (-x) = -lower x := by
...
+  rw [zsmul_eq_mul, Rat.intCast_eq_divInt, Rat.divInt_mul_divInt', one_mul,
+    Rat.den_mk, Rat.den_mk]
+  rw [if_neg (Nat.cast_ne_zero.mpr nz), if_neg (Nat.cast_ne_zero.mpr nz)]
+  apply Nat.div_le_div_left
+  · apply Nat.le_of_dvd (by simp [Int.gcd_pos_iff, nz])
+    exact Int.gcd_dvd_gcd_mul_left t b n
+  · simp [Int.gcd_pos_iff, nz]
+
+@[simp]
+theorem lower_add_self {x : Dyadic} (hx : x.1.den ≠ 1) :
```
vihdzp, line 334: I think these lemmas on `lower` and `upper` might be nice to have. They might help us if we ever want to work with dyadic games explicitly in the future.

plp127:
closing since we have this already

PR #109, opened 2025-07-03
feat: `toIGame (x + y) ≈ toIGame x + toIGame y`

vihdzp:
We show that dyadic game addition works as expected, and as a corollary, dyadic games are equivalent to their dyadic rationals.


CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -4,13 +4,13 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Violeta Hernández Palacios
 -/
 import CombinatorialGames.Game.Short
-import Mathlib.Analysis.Normed.Field.Lemmas
+import CombinatorialGames.Surreal.Division
```
plp127, line 7: If you're having to import division now you may want to consider splitting this file?

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -157,11 +157,19 @@ abbrev den (x : Dyadic) : ℕ := x.1.den
 theorem den_ne_zero (x : Dyadic) : x.den ≠ 0 := x.1.den_ne_zero
 theorem den_pos (x : Dyadic) : 0 < x.den := x.1.den_pos
 theorem den_mem_powers (x : Dyadic) : x.den ∈ Submonoid.powers 2 := x.2
+theorem one_le_den (x : Dyadic) : 1 ≤ x.den := x.den_pos
```
plp127, line 160: I don't like this defeq abuse, consider using `Nat.lt_iff_add_one_le`

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -318,10 +334,27 @@ theorem odd_num {x : Dyadic} (hx : x.den ≠ 1) : Odd x.num := by
   rw [← Int.natAbs_dvd_natAbs]
   exact (Nat.not_coprime_of_dvd_of_dvd one_lt_two · hd x.1.reduced)
 
-theorem num_eq_self_of_den_eq_one {x : Dyadic} (hx : x.den = 1) : x.num = x := by
+theorem coe_int_num_of_den_eq_one {x : Dyadic} (hx : x.den = 1) : x.num = x := by
```
plp127: I have no idea what `coe_int_num_of_den_eq_one` is supposed to mean, I think the name was better before.

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -334,6 +367,11 @@ theorem eq_mkRat_of_den_le {x : Dyadic} {n : ℕ} (h : x.den ≤ n) (hn : n ∈
 instance : CanLift Dyadic Int Int.cast (·.1.den = 1) where
   prf x hx := ⟨x.1.num, Dyadic.ext (x.1.den_eq_one_iff.mp hx)⟩
 
+theorem den_add_le {x y : Dyadic} (h : x.den ≤ y.den) : (x + y).den ≤ y.den := by
```
plp127: Can you get rid of the `h` assumption with a `wlog`?

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -542,4 +617,81 @@ theorem toIGame_equiv_toIGame {x y : Dyadic} : toIGame x ≈ toIGame y ↔ x = y
...
+  · rw [Fits, forall_leftMoves_add, forall_rightMoves_add]
+    refine ⟨⟨?_, ?_⟩, ⟨?_, ?_⟩⟩ <;> intro z hz
+    any_goals
+      obtain rfl := eq_lower_of_mem_leftMoves_toIGame hz
+      rw [← (toIGame_add_equiv _ _).le_congr_right]
+      simpa using lower_lt _
+    all_goals
+      obtain rfl := eq_upper_of_mem_rightMoves_toIGame hz
+      rw [← (toIGame_add_equiv _ _).le_congr_left]
+      simpa using lt_upper _
```
plp127, line 628: Can this be done with the `first` combinator? I'm not a huge fan of `any_goals`.

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -542,4 +617,81 @@ theorem toIGame_equiv_toIGame {x y : Dyadic} : toIGame x ≈ toIGame y ↔ x = y
...
+  · intro z hz
+    obtain rfl := eq_lower_of_mem_leftMoves_toIGame hz
+    rw [not_fits_iff]
+    left
+    obtain h | h | h := lt_trichotomy x.den y.den
+    on_goal 2 => by_cases hy : y.den = 1; simp_all
```
plp127: I'm sure there's a better way you can phrase this that doesn't involve me having to keep track of what goals are open.

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -542,4 +617,81 @@ theorem toIGame_equiv_toIGame {x y : Dyadic} : toIGame x ≈ toIGame y ↔ x = y
...
+      have hx := toIGame_of_den_ne_one (den_ne_one_of_den_lt h)
+      have : toIGame (upper x) ∈ (toIGame x).rightMoves := by rw [hx]; simp
+      rw [← (toIGame_add_equiv ..).le_congr_left, toIGame_le_toIGame, hx]
+      simpa using le_upper_add_of_den_ge h.le
+    all_goals
+      use toIGame x + toIGame (upper y)
+      have hy := toIGame_of_den_ne_one (by first | exact hy | exact den_ne_one_of_den_lt h)
+      have : toIGame (upper y) ∈ (toIGame y).rightMoves := by rw [hy]; simp
+      rw [← (toIGame_add_equiv ..).le_congr_left, toIGame_le_toIGame, hy]
+      simpa using le_upper_add_of_den_le h.le
```
plp127: You seem to be repeating the same argument as above. Any chance this could be factored out into a separate lemma?

plp127 (commented):
Good job! This looks like it might be a bit simpler than what I had.

CombinatorialGames/Surreal/Numeric.lean
tristan-f-r: I do think that `coe_int_num` is clearer here for a name since it indicates exactly what this theorem is doing. `self` is misleading because of the implicit coercion.

CombinatorialGames/Surreal/Numeric.lean
tristan-f-r: There's a lot of symmetry between these two lemmas, but there's also a lot of differences. At best, this could use a bunch of `let`/`set` declarations, but I think that would be messier than duplicating it here.

CombinatorialGames/Surreal/Numeric.lean
plp127: How about `intCast_num_eq_self_of_den_eq_one` or similar?

CombinatorialGames/Surreal/Numeric.lean
vihdzp: I don't think a split makes much sense - we need the division import to show `toIGame_equiv_ratCast`, which to me seems like a very important theorem you can't get much done without.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: I changed the name to match the name of the statement on `Rat`, but if you don't like that name we can keep it as is.

CombinatorialGames/Surreal/Numeric.lean
plp127: Can you give an example of something you can't do without ratcast?

CombinatorialGames/Surreal/Numeric.lean
vihdzp: The theorem isn't true without that assumption. We could rewrite it as `(x + y).den ≤ max x.den y.den` but that seems a bit less useful.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: I'm not 100% sure what you want me to replace the code by, but would it not just be a more verbose version of what's written here?

`any_goals` can sometimes have weird side effects, but it shouldn't here. The two goals it's not intended to solve should fail at the very first `obtain`.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: I'm not sure if I could. The basic idea here is that we case on whether `x.den` is less, equal, or greater than `y.den`, but also, we have to do the case `x.den = y.den = 1` separately or our induction doesn't work (that's what the `by_cases; simp_all` is doing). 

Once that's dealt with, the cases `x.den < y.den` and `x.den = y.den` are pretty much identical, which is why we first solve the last goal, and then do these other two at once.

CombinatorialGames/Surreal/Numeric.lean
plp127: Oh, I must have either misread or misinterpreted the theorem statement. Can this be something like `den_add_le_den_right`? I'm not sure what would make this harder to mistake.

CombinatorialGames/Surreal/Numeric.lean
plp127: In this case I think mathlib has the wrong name.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: Proving that `toIGame` preserves multiplication seems horrendously messy without going through rationals. Likewise, without this result, we can't do the often convenient trick of proving an inequality by passing to `Game` or `Surreal` and proving it there.

Besides, surreal division only really adds two files into the import list, and I imagine it will stay as one of our most basic imports in the folder.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: I'm usually one to be opposed to def-eq abuse, but since this is constrained to a very short proof I don't really think there's a problem.

CombinatorialGames/Surreal/Numeric.lean
plp127: I am thinking of an inductive proof right now, but I guess if you use rationals the proof is basically free.

CombinatorialGames/Surreal/Numeric.lean
vihdzp: Oh, nevermind! I realize I can simplify the proof by simply casing on whether `x.den ≤ y.den` or `y.den < x.den` instead; I don't need the full trichotomy.

CombinatorialGames/Surreal/Numeric.lean
```diff
@@ -536,4 +611,77 @@ theorem toIGame_equiv_toIGame {x y : Dyadic} : toIGame x ≈ toIGame y ↔ x = y
...
+      use toIGame x + toIGame (upper y)
+      have hy := toIGame_of_den_ne_one hy
+      have : toIGame (upper y) ∈ (toIGame y).rightMoves := by rw [hy]; simp
+      rw [← (toIGame_add_equiv ..).le_congr_left, toIGame_le_toIGame, hy]
+      simpa using le_upper_add_of_den_le h
+    · use toIGame (upper x) + toIGame y
+      have hx := toIGame_of_den_ne_one (den_ne_one_of_den_lt h)
+      have : toIGame (upper x) ∈ (toIGame x).rightMoves := by rw [hx]; simp
+      rw [← (toIGame_add_equiv ..).le_congr_left, toIGame_le_toIGame, hx]
+      simpa using le_upper_add_of_den_ge h.le
```
tristan-f-r, line 660:
```suggestion
      rw [← (toIGame_add_equiv ..).le_congr_left, hx]
```

tristan-f-r:
Oh! My bad - I commented on the wrong line :sweat:. The `toIGame_le_toIGame` from both `rw` calls can be dropped in `toIGame_add_equiv`.

tristan-f-r (approved):
[I would also remove `toIGame_le_toIGame` from the other side] Using the simplicity theorem to prove this is a good way of doing this! I'm not sure if it would be simpler to prove that addition over decrementing the denominator is preserved (2^(-n-1) + 2^(-n-1) = 2^(-n)) then generalizing that to arbitrary dyadics, but I think this proof is more motivated since dyadics are explicitly constructed with incrementing and decrementing the numerator anyways.

vihdzp:
I tried proving this by using `2^(-(n+1)) + 2^(-(n+1)) = 2^(-n)` to no avail. The fact that these games always have `0` as a left option simplifies reasoning about their sums quite a bit, and I wasn't able to generalize away from this.

PR #111, opened 2025-07-04
feat: define `IGame.toDyadic`

vihdzp:
Any short, numeric game is equivalent to some dyadic rational, which is given by the `IGame.toDyadic` function.

There's some interesting things we could do with this file, which I've outlined in the module docstring. I don't currently plan to do these, but if someone else wants to pursue them, go ahead!

Also, I fixed the file name. Not sure how I didn't catch that earlier.

- [x] Depends on: #112


tristan-f-r (approved):
`equiv_dyadic` is written very well. Also, good catch on the file name and the cmt rename 👍

PR #114, opened 2025-07-08
feat: surreal birthday

plp127:
Define the birthday of a surreal number as the least birthday among all *numeric* options defining it. Copy-paste some properties from the corresponding definition for games.

### DIFF
--- a/CombinatorialGames/Game/Birthday.lean
+++ b/CombinatorialGames/Game/Birthday.lean
@@ -430,7 +430,7 @@ theorem birthday_sub_le (x y : Game) : (x - y).birthday ≤ x.birthday + y.birth
 See https://mathoverflow.net/a/476829/147705. -/
 
 /-- Games with a bounded birthday form a small set. -/
-instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x // birthday x ≤ o} := by
+instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x | birthday x ≤ o} := by
   have : Small.{u} (mk '' {x | IGame.birthday x ≤ o}) :=
     @small_image _ _ _ _ (IGame.small_setOf_birthday_le o)
   refine @small_subset _ _ _ ?_ this
@@ -441,7 +441,7 @@ instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x // birthday
   simp_all
 
 /-- Games with a bounded birthday form a small set. -/
-instance small_setOf_birthday_lt (o : NatOrdinal.{u}) : Small.{u} {x // birthday x < o} := by
+instance small_setOf_birthday_lt (o : NatOrdinal.{u}) : Small.{u} {x | birthday x < o} := by
   apply @small_subset _ _ _ _ (small_setOf_birthday_le o)
   exact fun x (hx : x.birthday < _) ↦ le_of_lt hx
 
--- a/CombinatorialGames/Surreal/Birthday.lean
+++ b/CombinatorialGames/Surreal/Birthday.lean
@@ -1,0 +1,112 @@
+/-
+Copyright (c) 2025 Aaron Liu. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Aaron Liu
+-/
+import CombinatorialGames.Game.Birthday
+import CombinatorialGames.Surreal.Ordinal
+
+universe u
+noncomputable section
+
+namespace Surreal
+open IGame NatOrdinal Order Set
+
+def birthday (x : Surreal.{u}) : NatOrdinal.{u} :=
+  sInf (IGame.birthday '' {c | ∃ _ : Numeric c, mk c = x})
+
+theorem birthday_eq_iGameBirthday (x : Surreal) :
+    ∃ (y : IGame) (_ : Numeric y), mk y = x ∧ y.birthday = birthday x := by
+  have h : (IGame.birthday '' {c | ∃ _ : Numeric c, mk c = x}).Nonempty := by
+    rw [image_nonempty]
+    exact ⟨_, _, x.out_eq⟩
+  obtain ⟨c, ⟨cn, rfl⟩, hc⟩ := csInf_mem h
+  exact ⟨c, cn, rfl, hc⟩
+
+theorem birthday_mk_le (x : IGame) [Numeric x] : birthday (mk x) ≤ x.birthday :=
+  csInf_le' ⟨x, ⟨_, rfl⟩, rfl⟩
+
+@[simp]
+theorem birthday_zero : birthday 0 = 0 := by
+  simpa using birthday_mk_le 0
+
+@[simp]
+theorem birthday_eq_zero {x : Surreal} : birthday x = 0 ↔ x = 0 := by
+  obtain ⟨_, _, _, _⟩ := birthday_eq_iGameBirthday x
+  refine ⟨fun _ ↦ ?_, ?_⟩ <;> simp_all
+
+private theorem birthday_neg_le (x : Surreal) : (-x).birthday ≤ x.birthday := by
+  obtain ⟨y, _, hy, hy'⟩ := birthday_eq_iGameBirthday x
+  rw [← hy', ← hy]
+  apply (birthday_mk_le _).trans
+  rw [IGame.birthday_neg]
+
+@[simp]
+theorem birthday_neg (x : Surreal) : (-x).birthday = x.birthday := by
+  apply (birthday_neg_le x).antisymm
+  simpa using birthday_neg_le (-x)
+
+theorem le_toSurreal_birthday (x : Surreal) : x ≤ x.birthday.toSurreal := by
+  obtain ⟨y, _, hy, hy'⟩ := birthday_eq_iGameBirthday x
+  rw [← hy', ← hy]
+  exact y.le_toIGame_birthday
+
+theorem neg_toSurreal_birthday_le (x : Surreal) : -x.birthday.toSurreal ≤ x := by
+  simpa [neg_le] using le_toSurreal_birthday (-x)
+
+@[simp]
+theorem birthday_toSurreal (o : NatOrdinal) : birthday o.toSurreal = o := by
+  apply le_antisymm
+  · simpa using birthday_mk_le o.toIGame
+  · simpa using o.toSurreal.le_toSurreal_birthday
+
+@[simp, norm_cast]
+theorem birthday_natCast (n : ℕ) : birthday n = n := by
+  simpa using birthday_toSurreal n
+
+@[simp]
+theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] : birthday ofNat(n) = n :=
+  birthday_natCast n
+
+@[simp]
+theorem birthday_one : birthday 1 = 1 := by
+  simpa using birthday_natCast 1
+
+theorem birthday_add_le (x y : Surreal) : (x + y).birthday ≤ x.birthday + y.birthday := by
+  obtain ⟨a, _, ha, ha'⟩ := birthday_eq_iGameBirthday x
+  obtain ⟨b, _, hb, hb'⟩ := birthday_eq_iGameBirthday y
+  rw [← ha', ← hb', ← ha, ← hb, ← IGame.birthday_add]
+  exact birthday_mk_le _
+
+theorem birthday_sub_le (x y : Surreal) : (x - y).birthday ≤ x.birthday + y.birthday := by
+  simpa using birthday_add_le x (-y)
+
+/- The bound `(x * y).birthday ≤ x.birthday * y.birthday` on surreals is currently an open problem.
+See https://mathoverflow.net/a/476829/147705. -/
+
+/--
+The birthday of a surreal number is at least the birthday of the corresponding game.
+-/
+theorem birthday_toGame_le (x : Surreal) : x.toGame.birthday ≤ x.birthday := by
+  obtain ⟨c, _, rfl, h⟩ := birthday_eq_iGameBirthday x
+  rw [← h, toGame_mk]
+  exact Game.birthday_mk_le c
+
+-- See https://mathoverflow.net/q/497321
+proof_wanted birthday_toGame (x : Surreal) : x.toGame.birthday = x.birthday
+
+/-- Surreals with a bounded birthday form a small set. -/
+instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x | birthday x ≤ o} := by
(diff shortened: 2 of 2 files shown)


CombinatorialGames/Game/Birthday.lean
```diff
@@ -430,7 +430,7 @@ theorem birthday_sub_le (x y : Game) : (x - y).birthday ≤ x.birthday + y.birth
 See https://mathoverflow.net/a/476829/147705. -/
 
 /-- Games with a bounded birthday form a small set. -/
-instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x // birthday x ≤ o} := by
```
vihdzp: I think we should keep the version with `//`, since that's the simp-normal form.

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,115 @@
...
+universe u
+noncomputable section
+
+namespace Surreal
+open IGame NatOrdinal Order Set
+
+/--
+The birthday of a surreal number is defined as the least birthday
+among all *numeric* pre-games that define it. -/
+def birthday (x : Surreal.{u}) : NatOrdinal.{u} :=
```
vihdzp, line 17: I feel uncomfortable about having this definition until we can actually prove (or disprove!) its connection to `Game.birthday`.

vihdzp (changes requested):

CombinatorialGames/Game/Birthday.lean
plp127: Unfortunately, this causes instance synthesis to fail.

tristan-f-r: [Is that a problem here? I believe that doing manual synthesization is better than having a non simp-normal form instance.]

plp127: I have about four different places I would really not like to write it in manually (what's the point of it being an instance if you write it in manually every time), so I prefer to keep it this way. Would it be better if I defined a `birthdaySet` (analagous to `birthdayFinset`) and proved some lemmas about that instead?

plp127: PS. I think `birthdayFinset` should give you the games lt to a specific birthday (instead of le), since this way you get the most amount of finsets.

tristan-f-r: While `birthdayFinset`'s motivating construction was to prove that nat birthdays sit in finsets, having it be containment makes more sense in connection to born games - though I do agree that there should be an accompanying finset that is strictly the specified birthday (something like `born...` and `birthday...` for the containment and strict sets respectively)

CombinatorialGames/Surreal/Birthday.lean
plp127: I have a proof idea, I'll see if it works.

plp127: I have upgraded the proof idea to a proof, but it's a bit cumbersome to type into Lean so I'll probably do it tomorrow.

vihdzp: Can you at least write down the argument informally? I'm really interested in this!

CombinatorialGames/Game/Birthday.lean
vihdzp: `birthdayFinset` exists as an aid in a particular proof, and I don't really think we should be building much API around it. 

I think the easiest solution is to just add auxiliary instances for the non-normal simp form. Alternatively, we could write something like `birthday ⁻¹' Iio o`, which while cumbersome, shouldn't have this same issue.

plp127: What's the point of making it in simp-normal form if you're not calling `simp` with these anyways?

vihdzp: The point is that the instance won't stop being found just because some tactic like `simp_all` or `aesop` decided to simplify it.

plp127: I have never encountered a situation where that happens. I have encountered a situation where the instance is not found because `Set.Elem (setOf p)` is not defeq to `Subtype p`. It seems you have encountered it too, based on the proof of this instance having `@small_image` in it.

vihdzp: It can happen, and it's annoying when it happens. Why not just have both instances?

plp127: Sure, that works too I guess

plp127: I will split off another PR when I get home.

CombinatorialGames/Surreal/Birthday.lean
plp127: Lots of progress today, only one `sorry` left! Though I'm beginning to suspect my argument might be circular (I really hope it's not)

CombinatorialGames/Game/Birthday.lean
plp127: Now we have both forms :)

CombinatorialGames/Surreal/Birthday.lean
plp127: I've reduced down to laboriously checking sixteen cases by hand

vihdzp: That's incredible! Really excited to see your full proof.

vihdzp: Whatever happened to the 16 cases? I mean, I'm glad that wasn't necessary, but still.

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+
+universe u
+noncomputable section
+
+namespace Surreal
+open IGame NatOrdinal Order Set
+
+/--
+The birthday of a surreal number is defined as the least birthday
+among all *numeric* pre-games that define it. -/
```
vihdzp: Nitpicky, but I think this looks a bit nicer?
```suggestion
/-- The birthday of a surreal number is defined as the least birthday
among all *numeric* pre-games that define it. -/
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+def birthday (x : Surreal.{u}) : NatOrdinal.{u} :=
+  sInf (IGame.birthday '' {c | ∃ _ : Numeric c, mk c = x})
+
+theorem birthday_eq_iGameBirthday (x : Surreal) :
+    ∃ (y : IGame) (_ : Numeric y), mk y = x ∧ y.birthday = birthday x := by
+  have h : (IGame.birthday '' {c | ∃ _ : Numeric c, mk c = x}).Nonempty := by
+    rw [image_nonempty]
+    exact ⟨_, _, x.out_eq⟩
+  obtain ⟨c, ⟨cn, rfl⟩, hc⟩ := csInf_mem h
+  exact ⟨c, cn, rfl, hc⟩
```
vihdzp: If you're willing to slightly change the theorem declaration, you can golf this down to:
```suggestion
theorem birthday_eq_iGameBirthday (x : Surreal) :
    ∃ (y : IGame), (∃ (_ : Numeric y), mk y = x) ∧ y.birthday = birthday x := by
  refine csInf_mem (image_nonempty.2 ?_)
  exact ⟨_, _, x.out_eq⟩
```
If you don't want to do that, you can also do
```suggestion
theorem birthday_eq_iGameBirthday (x : Surreal) :
    ∃ (y : IGame) (_ : Numeric y), mk y = x ∧ y.birthday = birthday x := by
  simp_rw [exists_and_right]
  refine csInf_mem (image_nonempty.2 ?_)
  exact ⟨_, _, x.out_eq⟩
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+  csInf_le' ⟨x, ⟨_, rfl⟩, rfl⟩
+
+@[simp]
+theorem birthday_zero : birthday 0 = 0 := by
+  simpa using birthday_mk_le 0
+
+@[simp]
+theorem birthday_eq_zero {x : Surreal} : birthday x = 0 ↔ x = 0 := by
+  obtain ⟨_, _, _, _⟩ := birthday_eq_iGameBirthday x
+  refine ⟨fun _ ↦ ?_, ?_⟩ <;> simp_all
```
vihdzp, line 36:
```suggestion
  aesop
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+
+private theorem birthday_neg_le (x : Surreal) : (-x).birthday ≤ x.birthday := by
+  obtain ⟨y, _, hy, hy'⟩ := birthday_eq_iGameBirthday x
+  rw [← hy', ← hy]
+  apply (birthday_mk_le _).trans
+  rw [IGame.birthday_neg]
```
vihdzp:
```suggestion
  obtain ⟨y, _, rfl, hy'⟩ := birthday_eq_iGameBirthday x
  rw [← hy', ← IGame.birthday_neg]
  exact birthday_mk_le _
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+
+@[simp]
+theorem birthday_neg (x : Surreal) : (-x).birthday = x.birthday := by
+  apply (birthday_neg_le x).antisymm
+  simpa using birthday_neg_le (-x)
+
+theorem le_toSurreal_birthday (x : Surreal) : x ≤ x.birthday.toSurreal := by
+  obtain ⟨y, _, hy, hy'⟩ := birthday_eq_iGameBirthday x
+  rw [← hy', ← hy]
```
vihdzp:
```suggestion
  obtain ⟨y, _, rfl, hy'⟩ := birthday_eq_iGameBirthday x
  rw [← hy']
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+  obtain ⟨a, _, ha, ha'⟩ := birthday_eq_iGameBirthday x
+  obtain ⟨b, _, hb, hb'⟩ := birthday_eq_iGameBirthday y
+  rw [← ha', ← hb', ← ha, ← hb, ← IGame.birthday_add]
+  exact birthday_mk_le _
+
+theorem birthday_sub_le (x y : Surreal) : (x - y).birthday ≤ x.birthday + y.birthday := by
+  simpa using birthday_add_le x (-y)
+
+/- The bound `(x * y).birthday ≤ x.birthday * y.birthday` on surreals is currently an open problem.
+See https://mathoverflow.net/a/476829/147705. -/
```
vihdzp: We finally have the tools to actually *state* this problem!
```suggestion
/- This is currently an open problem, see https://mathoverflow.net/a/476829/147705. -/
proof_wanted birthday_mul_le (x y : Surreal) : (x * y).birthday ≤ x.birthday * y.birthday
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+
+/--
+The birthday of a surreal number is at least the birthday of the corresponding game.
+-/
```
vihdzp: Again, matter of taste, but perhaps?
```suggestion
/-- The birthday of a surreal number is at least the birthday of the corresponding game. -/
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+theorem birthday_toGame_le (x : Surreal) : x.toGame.birthday ≤ x.birthday := by
+  obtain ⟨c, _, rfl, h⟩ := birthday_eq_iGameBirthday x
+  rw [← h, toGame_mk]
+  exact Game.birthday_mk_le c
+
+-- See https://mathoverflow.net/q/497321
```
vihdzp: You might want to link to my answer instead ;)

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+proof_wanted birthday_toGame (x : Surreal) : x.toGame.birthday = x.birthday
+
+/-- Surreals with a bounded birthday form a small set. -/
+instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x | birthday x ≤ o} := by
+  have h₁ : {x | birthday x ≤ o} ⊆ toGame ⁻¹' {x | x.birthday ≤ o} := by
+    intro x hx
+    exact x.birthday_toGame_le.trans hx
+  have h₂ := Set.restrictPreimage_injective {x | x.birthday ≤ o} toGame.injective
+  have : Small.{u} (toGame ⁻¹' {x | x.birthday ≤ o}) := small_of_injective h₂
+  exact small_subset h₁
```
vihdzp, line 102: You can golf this down to three lines if you don't mind it looking much uglier.
```suggestion
instance small_setOf_birthday_le (o : NatOrdinal.{u}) : Small.{u} {x | birthday x ≤ o} :=
  @small_subset _ _ _ (fun x ↦ x.birthday_toGame_le.trans)
    (small_of_injective <| restrictPreimage_injective {x | x.birthday ≤ o} toGame.injective)
```

CombinatorialGames/Surreal/Birthday.lean
```diff
@@ -0,0 +1,123 @@
...
+
+/-- Surreals with a bounded birthday form a small set. -/
+instance small_setOf_birthday_lt (o : NatOrdinal.{u}) : Small.{u} {x | birthday x < o} := by
+  apply @small_subset _ _ _ _ (small_setOf_birthday_le o)
+  exact fun x (hx : x.birthday < _) ↦ le_of_lt hx
```
vihdzp: A similar remark applies here.
```suggestion
instance small_setOf_birthday_lt (o : NatOrdinal.{u}) : Small.{u} {x | birthday x < o} :=
  @small_subset _ _ _ (fun x (hx : x.birthday < _) ↦ le_of_lt hx) (small_setOf_birthday_le o)
```

vihdzp (commented):
Now that we actually know this definition is correct, I feel much more comfortable having this PR.

Left a few golfs. Some of these are really aggressive so feel free to cherry pick what you like.

CombinatorialGames/Surreal/Birthday.lean
plp127: 16 cases was too hard so I made an another type, called it `Rept`, and proved `auxAuxAux` which reduced it to three cases.

vihdzp:
@plp127 You think you can take a look at my reviews? I want to use this material to define `simplestBtwn`.

CombinatorialGames/Surreal/Birthday.lean
plp127: I think I want it to be readable.

PR #115, opened 2025-07-09
feat: surreal gaps

plp127:
Set up basic theory of surreal gaps and left and right sections of games

### DIFF
--- a/CombinatorialGames/Surreal/Dedekind.lean
+++ b/CombinatorialGames/Surreal/Dedekind.lean
@@ -9,6 +9,7 @@ import CombinatorialGames.Surreal.Basic
 # Dedekind cuts of Surreals
 
 TODO: write module docstring
+TODO: generalize to dedekind cuts of arbitrary preorders
 -/
 
 universe u
@@ -356,6 +357,14 @@ theorem leftGame_toGame (x : Surreal) : leftGame x.toGame = leftSurreal x := by
 theorem rightGame_toGame (x : Surreal) : rightGame x.toGame = rightSurreal x := by
   apply Cut.ext <;> apply Set.ext <;> simp
 
+theorem leftGame_lt_rightGame_iff {x : Game} :
+    leftGame x < rightGame x ↔ x ∈ range Surreal.toGame := by
+  constructor
+  · rintro ⟨y, hyl, hyr⟩
+    exact ⟨y, le_antisymm hyl hyr⟩
+  · rintro ⟨x, rfl⟩
+    simp [leftSurreal_lt_rightSurreal]
+
 /--
 Auxiliary definition for computing the `leftGame` of an explicitly given game.
 -/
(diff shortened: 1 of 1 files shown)


tristan-f-r (changes requested):
Preliminary note: while this codifies dedekind cuts with arbitrary Surreals, the 'surreal gaps' notion above is misleading: you would only be able to find gaps inside the Surreals if `Cut` took a proper class of Surreals.

Is this dedekind cut notion meant to formalize something that isn't surreal gaps? Or is it meant to formalize some other alternative construction of the Surreals? (One of my personal favorite ways of thinking about constructing the surreals is through this currently formalized notion of dedekind cuts over sets)

plp127:
> you would only be able to find gaps inside the Surreals if `Cut` took a proper class of Surreals

But it does??? Can you elaborate?

plp127:
> Is this dedekind cut notion meant to formalize something that isn't surreal gaps? Or is it meant to formalize some other alternative construction of the Surreals?

These are surreal gaps; for example `⊤` is the On gap and `Cut.leftSurreal 0` is the 1 / On gap

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+structure Cut where
+  /--
+  The lower set of a Dedekind section.
+  -/
+  left : Set Surreal
+  /--
+  The upper set of a Dedekind section.
+  -/
+  right : Set Surreal
+  left_lf_right' : ∀ l ∈ left, ∀ r ∈ right, l ⧏ r
```
vihdzp, line 32: Why not simply write `l < r`? Surreals are a total order, after all.

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+  codisjoint' : Codisjoint left right
+
+namespace Cut
+
+noncomputable instance : DecidableEq Cut := Classical.decEq Cut
```
vihdzp, line 37: What's the point of having a noncomputable `DecidableEq` instance?

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+  simpa using codisjoint_iff.1 c.codisjoint
+
+@[simp]
+theorem right_union_left {c : Cut} : c.right ∪ c.left = univ := by
+  simpa using codisjoint_iff.1 c.codisjoint.symm
+
+theorem disjoint {c : Cut} : Disjoint c.left c.right := by
+  rw [disjoint_iff_inter_eq_empty, eq_empty_iff_forall_notMem]
+  intro x hx
+  exact c.left_lf_right hx.left hx.right le_rfl
```
vihdzp, line 58: If you want to:
```suggestion
  exact fun x hx ↦ c.left_lf_right hx.left hx.right le_rfl
```

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+  rw [le_iff_right, ← compl_subset_compl, compl_right, compl_right]
+
+instance : LT Cut where
+  lt a b := (b.left ∩ a.right).Nonempty
+
+noncomputable instance : DecidableLT Cut := Classical.decRel LT.lt
+
+theorem lt_iff_nonempty {a b : Cut} : a < b ↔ (b.left ∩ a.right).Nonempty := Iff.rfl
+
+protected theorem not_le {a b : Cut} : ¬a ≤ b ↔ b < a := by
```
vihdzp, line 116: This and `le_of_lt` should be private, or better yet, inlined into the order instances.

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+
+theorem le_iff_left {a b : Cut} : a ≤ b ↔ a.left ⊆ b.left := by
+  rw [le_iff_right, ← compl_subset_compl, compl_right, compl_right]
+
+instance : LT Cut where
+  lt a b := (b.left ∩ a.right).Nonempty
+
+noncomputable instance : DecidableLT Cut := Classical.decRel LT.lt
+
+theorem lt_iff_nonempty {a b : Cut} : a < b ↔ (b.left ∩ a.right).Nonempty := Iff.rfl
```
vihdzp, line 114:
```suggestion
theorem lt_iff_nonempty {a b : Cut} : a < b ↔ (b.left ∩ a.right).Nonempty := .rfl
```

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+  simp_rw [le_iff_left, not_subset, ← mem_compl_iff]
+  rw [← inter_nonempty_iff_exists_left, compl_left, lt_iff_nonempty]
+
+protected theorem le_of_lt {a b : Cut} (h : a < b) : a ≤ b := by
+  rw [lt_iff_nonempty] at h
+  rw [le_iff_left]
+  obtain ⟨x, hx⟩ := h
+  intro y hy
+  refine b.isLowerSet_left ?_ hx.left
+  exact le_of_not_ge (a.left_lf_right hy hx.right)
```
vihdzp, line 126: You can merge this into a single line if you want
```suggestion
  exact b.isLowerSet_left (fun y hy ↦ le_of_not_ge (a.left_lf_right hy hx.right)) hx.left
```

CombinatorialGames/Surreal/Dedekind.lean
```diff
@@ -0,0 +1,450 @@
...
+    left_lf_right' := by simp
+    codisjoint' _ h _ := by simpa using h
+  }
+  le_top c := ⟨subset_univ c.left, empty_subset c.right⟩
+  bot := {
+    left := ∅
+    right := univ
+    left_lf_right' := by simp
+    codisjoint' _ _ h := by simpa using h
+  }
```
vihdzp, line 152:
```suggestion
  top := ⟨univ, ∅, by simp, by simpa using h⟩
  le_top c := ⟨subset_univ c.left, empty_subset c.right⟩
  bot := ⟨∅, univ, by simp, fun _ _ h ↦ by simpa using h⟩
```

vihdzp (commented):
Some initial remarks.

A note about this whole PR: every preorder can be embedded in a complete lattice via the Dedekind-MacNeille completion. I think it'd be better to develop this and PR it to Mathlib, rather than just doing this for surreal numbers.

vihdzp:
Oh, it turns out we already do have the Dedekind completion in Mathlib! You can write it as [`Concept`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Concept.html#Concept)`Surreal Surreal (· ≤ ·)`.

CombinatorialGames/Surreal/Dedekind.lean
plp127: Same reason we have `Real.decidableEq` in mathlib (and same for the le and lt instances)

CombinatorialGames/Surreal/Dedekind.lean
plp127: You seem to have dropped a binder somewhere, `h` is unbound now.

CombinatorialGames/Surreal/Dedekind.lean
plp127: I was considering what this should be for non-total orders, and I thought I should make it this to mimic the case for a general preorder.

CombinatorialGames/Surreal/Dedekind.lean
plp127: Why? It's not like `Nat.le_of_lt` is private. I've decided to not inline these because I use them twice.

plp127:
I think I should clarify: what I was looking for with my definition was not a way to embed an order into a complete lattice, but a way to characterize the *cuts* inside that order, where a cut is a partition of the order into a lower set and an upper set. I think that's what I accomplished here.

vihdzp:
Oh, I should write down what we discussed on DMs. It turns out this *is* the correct notion of gaps; that is, surreal gaps are not the same as the Dedekind completion, and every surreal actually corresponds to two gaps instead of one. Still, it's possible to simplify this treatment by defining `Gap = Concept Surreal Surreal (· < ·)` (using the `<` relation instead of the `≤` relation, which would give the Dedekind completion).

CombinatorialGames/Surreal/Dedekind.lean
vihdzp: `Nat.le_of_lt` isn't private because a lot of basic theorems on naturals are developed before the corresponding theorems on orders are! Every other type shouldn't have to do this.

plp127:
I'll close this now since #125 got merged.

PR #117, opened 2025-07-10
fix: delab antisymmrel and incomprel (again)

plp127:
Fail over to the default elaborator if the relation is not defeq to LE.

This *should* be the last PR fixing the delab.


plp127:
Is there a place to put tests, so I can test these and make sure that they're not still broken in some obscure way?

vihdzp (approved):
Sure, seems reasonable

plp127:
I can write some tests, if you can figure out how to get CI to run them.

PR #127, opened 2025-07-15
feat: more basic results on surreal cuts

vihdzp:
Nothing very mathematical yet, just a bunch of extra simp lemmas and a new `Neg Cut` instance.

### DIFF
--- a/CombinatorialGames/Surreal/Cut.lean
+++ b/CombinatorialGames/Surreal/Cut.lean
@@ -52,6 +52,9 @@ alias isCompl_left_right := Concept.isCompl_extent_intent
 alias isLowerSet_left := Concept.isLowerSet_extent'
 alias isUpperSet_left := Concept.isUpperSet_intent'
 
+@[ext] theorem ext {c d : Cut} (h : c.left = d.left) : c = d := Concept.ext h
+theorem ext' {c d : Cut} (h : c.right = d.right) : c = d := Concept.ext' h
+
 @[simp] theorem left_subset_left_iff {c d : Cut}: c.left ⊆ d.left ↔ c ≤ d := .rfl
 @[simp] theorem left_ssubset_left_iff {c d : Cut} : c.left ⊂ d.left ↔ c < d := .rfl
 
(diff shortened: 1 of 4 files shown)


CombinatorialGames/Surreal/Cut.lean
```diff
@@ -52,6 +52,9 @@ alias isCompl_left_right := Concept.isCompl_extent_intent
 alias isLowerSet_left := Concept.isLowerSet_extent'
 alias isUpperSet_left := Concept.isUpperSet_intent'
 
+@[ext] theorem ext {c d : Cut} (h : c.left = d.left) : c = d := Concept.ext h
+theorem ext' {c d : Cut} (h : c.right = d.right) : c = d := Concept.ext' h
```
plp127, line 56: Can we have an `ext_iff : c = d ↔ c.left = d.left ∧ c.right = d.right`? And maybe a `left_eq_left_iff` and `right_eq_right_iff` too.

CombinatorialGames/Surreal/Cut.lean
```diff
@@ -60,6 +63,15 @@ alias isUpperSet_left := Concept.isUpperSet_intent'
 @[simp] theorem right_ssubset_right_iff {c d : Cut} : c.right ⊂ d.right ↔ d < c :=
   Concept.intent_ssubset_intent_iff
```
plp127, line 70: Similarly for subset too, I think I want a `c ≤ d ↔ c.left ⊆ d.left ∧ d.right ⊆ c.right`.

CombinatorialGames/Surreal/Cut.lean
vihdzp: I don't really see the point of an `ext_iff` with both assumptions at once. The implication is trivial, and the reverse implication has redundant hypotheses.

vihdzp: I added the `left_inj` and `right_inj` theorems.

tristan-f-r (approved):

PR #129, opened 2025-07-15
feat: simplest surreal between two cuts

vihdzp:
We define the `simplestBtwn` function returning the simplest surreal between two cuts `x < y`, and relate it to cuts of games.

Co-authored by @plp127.

- [x] Depends on: #128

PR #130, opened 2025-07-15
feat: cuts of games are "small"

vihdzp:
We define small cuts, and prove that the left/right cuts of a game are always small.

For a definition of small cuts, see my [Math Overflow answer](https://mathoverflow.net/a/497645/147705).

- [ ] Depends on: #129

### DIFF
--- a/CombinatorialGames/Surreal/Cut.lean
+++ b/CombinatorialGames/Surreal/Cut.lean
@@ -518,11 +518,9 @@ private theorem small_game (x : IGame) : (leftGame (.mk x)).Small ∧ (rightGame
   obtain h | h := lt_or_ge (supLeft x) (infRight x)
   · rw [← simplestBtwn_supLeft_infRight h]
     simp
-  · rw [leftGame_eq_supLeft_of_le h, rightGame_eq_infRight_of_le h, supLeft, infRight,
-      iSup_subtype', iInf_subtype']
-    constructor
-    · exact .iSup fun ⟨y, hy⟩ ↦ (small_game _).2
-    · exact .iInf fun ⟨y, hy⟩ ↦ (small_game _).1
+  · rw [leftGame_eq_supLeft_of_le h, rightGame_eq_infRight_of_le h,
+      supLeft, infRight, iSup_subtype', iInf_subtype']
+    exact ⟨.iSup fun _ ↦ (small_game _).2, .iInf fun _ ↦ (small_game _).1⟩
 termination_by x
 decreasing_by igame_wf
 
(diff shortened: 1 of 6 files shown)


CombinatorialGames/Surreal/Cut.lean
```diff
@@ -149,16 +247,295 @@ theorem mem_right_rightSurreal {x y} : y ∈ (rightSurreal x).right ↔ x < y :=
...
+
+/-- A "small cut" is defined as either the infimum of a small set of left cuts of surreals,
+or the supremum of a small set of right cuts of surreals.
+
+Equivalently, small cuts are the closure of left and right cuts of surreals under small infima and
+suprema.
+
+This isn't a term in the literature, but it's useful for proving that birthdays of surreals equal
+those of their associated games. -/
+protected inductive Small : Cut.{u} → Prop
```
plp127: I don't really like `Small`, can this be `IsSmall` instead?

vihdzp: Sure, I agree.

CombinatorialGames/Surreal/Cut.lean
```diff
@@ -149,16 +247,298 @@ theorem mem_right_rightSurreal {x y} : y ∈ (rightSurreal x).right ↔ x < y :=
...
+  · rw [← compl_left]
+    intro z hz _ hz'
+    apply (hy' ▸ birthday_lt_of_mem_rightMoves hz).not_ge
+    apply (birthday_simplestBtwn_le_of_mem ..).trans (birthday_mk_le z)
+    refine ⟨isUpperSet_right _ (mk_lt_mk.2 <| Numeric.lt_rightMove hz).le ?_, hz'⟩
+    aesop
+
+/-! ### Small cuts -/
+
+/-- A "small cut" is defined as either the infimum of a small set of left cuts of surreals,
```
tristan-f-r, line 437: Should this be a class? It doesn't help here, but it would be consistent how this is an inductively defined proposition whose prefixed with `Is`

```suggestion
class inductive IsSmall : Cut.{u} → Prop
  | sInf' (s : Set Surreal) [Small.{u} s] : (sInf (leftSurreal '' s)).IsSmall
  | sSup' (s : Set Surreal) [Small.{u} s] : (sSup (rightSurreal '' s)).IsSmall
```

CombinatorialGames/Surreal/Cut.lean
tristan-f-r: Should `IsSmall` also be a class?

vihdzp: It seems potentially helpful for it to be, I'll try that.

PR #132, opened 2025-07-16
feat: birthday of surreal number corresponds to birthday of its game

vihdzp:
This proof was originally found by @plp127 in #123; I've simplified it somewhat by introducing a new definition `Cut.birthday`.

PR #134, opened 2025-07-17
feat: (leftSurreal x).birthday = x.birthday + 1

vihdzp:
Cool lemma that wasn't needed for #132 but which I managed to prove anyways.

A consequence of this and `birthday_iInf_le` is that the restriction to infima of only left cuts or suprema of only right cuts in the definition of `Cut.birthday` is actually unnecessary (even if helpful!)

- [x] Depends on: #132


CombinatorialGames/Surreal/Cut.lean
```diff
@@ -288,7 +309,33 @@ theorem rightSurreal_le_iff {x : Surreal} {y : Cut} : rightSurreal x ≤ y ↔ x
...
+theorem rightSurreal_lt_leftSurreal_iff {x y : Surreal} :
+    rightSurreal x < leftSurreal y ↔ x < y := by
+  rw [← left_ssubset_left_iff, left_rightSurreal, left_leftSurreal]
+  constructor <;> intro h
+  · exact h.1 (mem_Iic.2 le_rfl)
+  · constructor <;> simpa
+
+theorem leftSurreal_covBy_rightSurreal (x : Surreal) : leftSurreal x ⋖ rightSurreal x := by
+  refine ⟨leftSurreal_lt_rightSurreal x, fun y ↦ ?_⟩
+  simp [← mem_compl_iff]
```
plp127: Should we have some `@[simp]` lemmas `notMem_left` and `notMem_right`?

vihdzp: That's a really good idea, I'll add them to the PR.

tristan-f-r (approved):

PR #137, opened 2025-07-18
chore(game_cmp): nat decidability

tristan-f-r:
`game_cmp` can be fast too, if we allow it to decide on ℕ. (This is as strong as adding the same lemmas on ℤ)

### DIFF
--- a/CombinatorialGames/Game/Basic.lean
+++ b/CombinatorialGames/Game/Basic.lean
@@ -248,11 +248,11 @@ theorem sub_mul_equiv (x y z : IGame) : (x - y) * z ≈ x * z - y * z :=
 theorem mul_assoc_equiv (x y z : IGame) : x * y * z ≈ x * (y * z) :=
   Game.mk_eq_mk.1 (Game.mk_mul_assoc x y z)
 
-@[simp, norm_cast]
+@[simp, norm_cast, game_cmp]
 theorem natCast_le {m n : ℕ} : (m : IGame) ≤ n ↔ m ≤ n := by
   simp [← Game.mk_le_mk]
 
-@[simp, norm_cast]
+@[simp, norm_cast, game_cmp]
 theorem natCast_lt {m n : ℕ} : (m : IGame) < n ↔ m < n := by
   simp [← Game.mk_lt_mk]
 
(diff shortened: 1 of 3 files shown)


CombinatorialGames/Game/Tactic.lean
```diff
@@ -40,7 +40,8 @@ macro "game_cmp" : tactic =>
     try simp only [lt_iff_le_not_ge, ge_iff_le, gt_iff_lt, AntisymmRel, CompRel, IncompRel]
     repeat
       rw [IGame.le_iff_forall_lf]
-      simp only [game_cmp]})
+      simp only [game_cmp]
+      try decide})
```
vihdzp, line 44: It makes me a bit paranoid to see `decide` tried on every loop iteration. Do we know for a fact `decide` will always fail fast on any goal not of the form of a natural number comparison?

An alternate approach might be to instead tag the simp-proc for nat comparison with `game_cmp`. That way it will fire when it sees a natural number comparison, and won't fire for any goals it won't be able to solve.

CombinatorialGames/Game/Basic.lean
```diff
@@ -248,11 +248,11 @@ theorem sub_mul_equiv (x y z : IGame) : (x - y) * z ≈ x * z - y * z :=
 theorem mul_assoc_equiv (x y z : IGame) : x * y * z ≈ x * (y * z) :=
   Game.mk_eq_mk.1 (Game.mk_mul_assoc x y z)
 
-@[simp, norm_cast]
+@[simp, norm_cast, game_cmp]
 theorem natCast_le {m n : ℕ} : (m : IGame) ≤ n ↔ m ≤ n := by
   simp [← Game.mk_le_mk]
 
-@[simp, norm_cast]
+@[simp, norm_cast, game_cmp]
```
vihdzp, line 255: This `game_cmp` does nothing, since the tactic converts `<` to `≤` and `⧏` before it even starts applying simp lemmas.

CombinatorialGames/Test/GameCmp.lean
```diff
@@ -27,6 +27,11 @@ example : (2 : IGame) + 2 ≈ 4 := by game_cmp
 example : (3 : IGame) - 2 ≈ 1 := by game_cmp
 example : (2 : IGame) * 2 ≈ 4 := by game_cmp
 
+-- Int/Nat decidability
+example : (200 : IGame) < 400 := by game_cmp
+example : (-200 : IGame) ≤ 400 := by game_cmp
+example : (-200 : IGame) ≈ -200 := by game_cmp
```
vihdzp, line 33: These tests are also a bit misleading, they work just fine without even having to rely on decidability! (The top one does time out currently)

tristan-f-r:
Doing this through simprocs is quite bad as well. Closing this.

PR #139, opened 2025-07-18
feat: product of impartial games is impartial

vihdzp:
Note that this would have been trivial had we required identity in the definition of `Impartial` games. The argument here is AFAIK original, improvements welcome!

- [x] depends on: #138


tristan-f-r (approved):
`mul'` can be squeezed a little more (#141), but that's all the comments from me!

PR #141, opened 2025-07-24
chore: golf

tristan-f-r:
attached review PR since some of the golfs depend on the slightly changed def. This can be closed and added manually, or merged into #139.


vihdzp:
Nice!

PR #142, opened 2025-07-24
refactor: redefine `Real.toIGame` using `Dyadic` numbers

vihdzp:
Defining real numbers as Dedekind cuts of dyadic rationals rather than all rationals is a bit more mathematically elegant, and it allows us to state the theorem `(Real.toIGame x).birthday ≤ ω`.

A few auxiliary theorems were re-stated in terms of `Dyadic` rather than `ℚ`, but overall I wasn't able to find any major simplifications in arguments by using one or the other.


CombinatorialGames/Surreal/Real.lean
```diff
@@ -26,19 +27,36 @@ open IGame
...
-  simp_all [← Set.image_neg_eq_neg, Set.image_image]
+-- TODO: PR to Mathlib
+theorem exists_div_btwn {K : Type*} [Field K] [LinearOrder K] [IsStrictOrderedRing K]
+    [Archimedean K] {x y : K} {n : ℕ} (h : x < y) (nh : (y - x)⁻¹ < n) :
+    ∃ z : ℤ, x < (z : K) / n ∧ (z : K) / n < y := by
+  obtain ⟨z, zh⟩ := exists_floor (x * n)
+  refine ⟨z + 1, ?_⟩
+  have n0' := (inv_pos.2 (sub_pos.2 h)).trans nh
+  have n0 := Nat.cast_pos.1 n0'
+  rw [div_lt_iff₀ n0']
```
tristan-f-r:
```suggestion
  have n0' := (inv_pos.2 (sub_pos.2 h)).trans nh
  rw [div_lt_iff₀ n0']
```

CombinatorialGames/Surreal/Real.lean
```diff
@@ -195,12 +247,19 @@ theorem toIGame_add_ratCast_equiv (x : ℝ) (q : ℚ) : toIGame (x + q) ≈ x +
...
 
+theorem toIGame_add_dyadic_equiv (x : ℝ) (q : Dyadic) : toIGame (x + q) ≈ x + q :=
+  (toIGame_add_ratCast_equiv _ _).trans (add_congr_right q.toIGame_equiv_ratCast.symm)
+
+theorem toIGame_dyadic_add_equiv (q : Dyadic) (x : ℝ) : toIGame (q + x) ≈ q + x := by
+  simpa [add_comm] using toIGame_add_dyadic_equiv x q
+
 theorem toIGame_add_equiv (x y : ℝ) : toIGame (x + y) ≈ x + y := by
   rw [AntisymmRel, le_iff_forall_lf, le_iff_forall_lf, forall_leftMoves_add, forall_rightMoves_add]
   simp_rw [forall_leftMoves_toIGame, forall_rightMoves_toIGame, Numeric.not_le]
```
tristan-f-r, line 264: (`forall_leftMoves_toIGame` and `forall_rightMoves_toIGame` can be moved to `rw`. However, this is a little less nice - is that why we avoid it here?)

tristan-f-r (approved):

CombinatorialGames/Surreal/Real.lean
vihdzp: That was the idea, yeah. We can trim one line off the code by doing this.

PR #151, opened 2025-07-26
feat: lawnmower theorem on impartial games

vihdzp:
Impartial games are not necessarily dicotic, but they're always equivalent to a game of nim, which is.

- [ ] depends on: #150


CombinatorialGames/Game/Specific/Nim.lean
```diff
@@ -152,6 +153,11 @@ protected instance Impartial.nim (o : Nimber) : Impartial (nim o) := by
     exact .nim a
 termination_by o
 
+protected instance Dicotic.nim (o : Nimber) : Dicotic (nim o) := by
+  rw [dicotic_def]
+  simpa using fun a ha ↦ Dicotic.nim a
```
tristan-f-r:
```suggestion
  simpa using fun a ha ↦ .nim a
```

tristan-f-r (approved):

PR #153, opened 2025-07-27
feat: simplest extension theorem 1/4

vihdzp:
We prove that the nimbers that are groups are exactly the powers of two, and characterize arithmetic on them.
- [x] depends on: #155


CombinatorialGames/Nimber/Simplicity.lean
```diff
@@ -0,0 +1,288 @@
...
+  h.toIsRing.mul_lt hy (h.inv_lt hz)
+
+/-! ### Algebraically closed fields -/
+
+/-- A nimber `x` is algebraically closed when `IsField x`, and every polynomial in the nimbers with
+coefficients less than `x` has a root that's less than `x`. Note that `0` is algebraically closed
+under this definition. -/
+@[mk_iff]
+structure IsAlgClosed (x : Nimber) extends IsRing x where
+  has_root {p : Nimber[X]} (hp : ∀ n, p.coeff n < x) : ∃ r < x, p.IsRoot r
```
DjangoPeeters: I think this doesn't work for non-zero constant polynomials.

DjangoPeeters (changes requested):
Most of this looks fine to me. Apart from a definition, I don't have any other comments.

CombinatorialGames/Nimber/Simplicity.lean
vihdzp: You're right, I forgot a `p.degree ≠ 0` assumption

CombinatorialGames/Nimber/Basic.lean
```diff
@@ -74,7 +74,11 @@ namespace Nimber
 open Ordinal
 
 @[simp] theorem toOrdinal_symm_eq : Nimber.toOrdinal.symm = Ordinal.toNimber := rfl
-@[simp] theorem toOrdinal_toNimber (a : Nimber) : ∗(Nimber.toOrdinal a) = a := rfl
+@[simp] theorem toNimber_toOrdinal (a : Nimber) : ∗(Nimber.toOrdinal a) = a := rfl
+
+theorem toOrdinal_le_iff (a : Nimber) (b : Ordinal) : toOrdinal a ≤ b ↔ a ≤ ∗b := .rfl
+theorem toOrdinal_lt_iff (a : Nimber) (b : Ordinal) : toOrdinal a < b ↔ a < ∗b := .rfl
+theorem toOrdinal_eq_iff (a : Nimber) (b : Ordinal) : toOrdinal a = b ↔ a = ∗b := .rfl
```
tristan-f-r, line 81: Not sure where this applies since mathlib doesn't seem to be consistent about this, but as aforementioned in #146, you may want to drop `_iff` here.

CombinatorialGames/Nimber/Basic.lean
```diff
@@ -158,7 +162,11 @@ open Nimber
 namespace Ordinal
 
 @[simp] theorem toNimber_symm_eq : toNimber.symm = Nimber.toOrdinal := rfl
-@[simp] theorem toNimber_toOrdinal (a : Ordinal) : Nimber.toOrdinal (∗a) = a := rfl
+@[simp] theorem toOrdinal_toNimber (a : Ordinal) : Nimber.toOrdinal (∗a) = a := rfl
+
+theorem toNimber_le_iff (a : Ordinal) (b : Nimber) : ∗a ≤ b ↔ a ≤ b.toOrdinal := .rfl
+theorem toNimber_lt_iff (a : Ordinal) (b : Nimber) : ∗a < b ↔ a < b.toOrdinal := .rfl
+theorem toNimber_eq_iff (a : Ordinal) (b : Nimber) : ∗a = b ↔ a = b.toOrdinal := .rfl
```
tristan-f-r, line 169: Also drop `_iff` here if reasonable.

CombinatorialGames/Nimber/Simplicity.lean
```diff
@@ -0,0 +1,289 @@
...
+  rw [← one_add_one_eq_two, mul_add, mul_one]
+
+theorem lt_mul_iff {a b c : Ordinal} : a < b * c ↔ ∃ q < c, ∃ r < b, a = b * q + r := by
+  obtain rfl | hb₀ := eq_or_ne b 0; simp
+  refine ⟨fun h ↦ ⟨_, (Ordinal.div_lt hb₀).2 h, _, mod_lt a hb₀, (div_add_mod ..).symm⟩, ?_⟩
+  rintro ⟨q, hq, r, hr, rfl⟩
+  apply (add_left_strictMono hr).trans_le
+  simp_rw [← mul_succ]
+  apply mul_le_mul_left'
+  simpa
```
tristan-f-r, line 102: `simpa` seems to be being used in your recent PRs as a kind of embedded `exact?` - is this preferred in mathlib?

```suggestion
  exact Order.succ_le_iff.mpr hq
```

CombinatorialGames/Nimber/Simplicity.lean
```diff
@@ -0,0 +1,289 @@
...
+/-! ### Groups -/
+
+/-- Add two nimbers as ordinal numbers. -/
+scoped notation:65 x:arg "+ₒ" y:arg => ∗(toOrdinal x + toOrdinal y)
+
+/-- A nimber `x` is a group when `Iio x` is closed under addition. Note that `0` is a group under
+this definition. -/
+@[mk_iff]
+structure IsGroup (x : Nimber) where
+  add_lt {y z} (hy : y < x) (hz : z < x) : y + z < x
```
tristan-f-r, line 136: You could use a strict implicit binder here, which gets rid of the future explicit notation `refine ⟨@fun y z hy hz ↦ ?_⟩` at the cost of needing to explicitly enumerate `add_lt`'s explicit arguments. I've yet to see explicit binders be (non-obviously) avoided for struct hypothesis situations, so take this suggestion with some suspicion.

```suggestion
  add_lt ⦃y z⦄ (hy : y < x) (hz : z < x) : y + z < x
```

CombinatorialGames/Nimber/Simplicity.lean
```diff
@@ -0,0 +1,289 @@
...
+  induction y with | h y =>
+  induction z with | h z =>
+  obtain rfl | hy₀ := eq_or_ne y 0; simpa
+  obtain rfl | hz₀ := eq_or_ne z 0; simpa
+  have hy' : log 2 y < x := by rwa [← lt_opow_iff_log_lt one_lt_two hy₀]
+  have hz' : log 2 z < x := by rwa [← lt_opow_iff_log_lt one_lt_two hz₀]
+  have hm {x y : Ordinal} : x % 2 ^ y < 2 ^ y := mod_lt _ (opow_ne_zero _ two_ne_zero)
+  rw [← two_opow_log_add hy₀, ← two_opow_log_add hz₀,
+    (two_opow _).add_eq_of_lt' hm, (two_opow _).add_eq_of_lt' hm]
+  -- TODO: it'd be nicer to use `wlog` here, but it breaks the termination checker.
```
tristan-f-r, line 206: (thanks for this comment!)

tristan-f-r:
(Oh! This was already merged)

vihdzp:
Oh, oops! I didn't know you were going to review this too. I'll add these suggestions in a separate PR.

CombinatorialGames/Nimber/Basic.lean
vihdzp: I'm not completely sure about dropping `_iff` here. I think the lemma name `toOrdinal_le` could be misinterpreted as `a ≤ ∗b → toOrdinal a ≤ b`. (Arguably the same could be said in #146, but I think for lemmas with a clear simplest direction it's more obvious that you're talking about an `iff` lemma).

CombinatorialGames/Nimber/Simplicity.lean
vihdzp: `exact` is certainly preferred in situations like this, I think I just forgot to clean that one up.

CombinatorialGames/Nimber/Simplicity.lean
vihdzp: I like this. I've even used it for very similar situations in the past, see [`Ordinal.Principal`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/Principal.html#Ordinal.Principal), which is basically the same definition!

PR #159, opened 2025-07-28
feat: `Small {y | Subposition y x}`

vihdzp:
We can prove this very easily using our existing results on birthdays. However, we opt to go the long way round and prove the more general result
```lean 4
theorem small_setOf_transGen' [∀ x, Small.{u} {y // r y x}] (x : α) :
    Small.{u} {y // Relation.TransGen r y x} := sorry
```
This also gives us the theorem for free in the case of loopy games.


CombinatorialGames/Mathlib/Small.lean
```diff
@@ -0,0 +1,46 @@
...
+    · have := small_level x n
+      exact small_image ..
+    · simp_all
+
+private theorem small_sUnion_level (x : α) : Small.{u} (⋃₀ range (level r x)) := by
+  refine @small_sUnion _ _ ?_ ?_
+  · exact small_range ..
+  · simp [small_level]
+
+theorem small_setOf_transGen (x : α) : Small.{u} {y // Relation.TransGen r x y} := by
```
plp127: This theorem doesn't mention `setOf` in its statement

vihdzp: True. I don't think there'll be any ambiguity if I simply call this `small_transGen`.

CombinatorialGames/Mathlib/Small.lean
```diff
@@ -0,0 +1,46 @@
...
+  induction hy with
+  | single =>
+    use 1
+    simpa [level]
+  | tail hy hr IH =>
+    obtain ⟨n, hn⟩ := IH
+    use n + 1
+    rw [level]
+    simp_rw [mem_sUnion, mem_image, exists_exists_and_eq_and]
+    exact ⟨_, hn, hr⟩
```
tristan-f-r:
```suggestion
    simpa [level] using ⟨_, hn, hr⟩
```

PR #162, opened 2025-07-30
feat: set up `shake` command

vihdzp:
We set up a `noshake.json` file so that `lake exe shake` can work, and fix the few redundant imports that we had.


tristan-f-r (approved):
I didn't know about this!

PR #164, opened 2025-07-30
feat: simplest extension theorem part 2.5/4

vihdzp:
We prove that the properties of being a group/ring/field/algebraically closed field are closed under suprema.

I'm not sure this is actually needed for the full proof, but it's an interesting result nonetheless!

- [x] depends on: #161


CombinatorialGames/Nimber/Simplicity.lean
```diff
@@ -261,14 +327,25 @@ is a ring under this definition. -/
...
   __ := IsGroup.zero
 
+@[simp]
 theorem IsRing.one : IsRing 1 where
   mul_lt := by simp
   __ := IsGroup.one
 
+protected theorem IsRing.sSup {s : Set Nimber} (H : ∀ x ∈ s, IsRing x) :
+    IsRing (sSup s) :=
+  ⟨IsGroup.sSup fun x hx ↦ (H x hx).toIsGroup, Principal.sSup (fun x hx ↦ (H x hx).mul_lt)⟩
```
tristan-f-r:
```suggestion
  ⟨IsGroup.sSup fun x hx ↦ (H x hx).toIsGroup, Principal.sSup fun x hx ↦ (H x hx).mul_lt⟩
```

tristan-f-r (approved):

PR #168, opened 2025-08-05
feat: multiplication of loopy games

vihdzp:
Just as with well-founded games, we can uniquely characterize multiplication of loopy games through the theorems:

```lean 4
def mulOption (x y a b : LGame) : LGame :=
  a * y + x * b - a * b

theorem leftMoves_mul (x y : LGame) :
    (x * y).leftMoves = (fun a ↦ mulOption x y a.1 a.2) ''
      (x.leftMoves ×ˢ y.leftMoves ∪ x.rightMoves ×ˢ y.rightMoves) := sorry

theorem rightMoves_mul (x y : LGame) :
    (x * y).rightMoves = (fun a ↦ mulOption x y a.1 a.2) ''
      (x.leftMoves ×ˢ y.rightMoves ∪ x.rightMoves ×ˢ y.leftMoves) := sorry
```

To actually construct a function satisfying these properties is another thing entirely. For this, we make use of the observation that any subposition of `x * y` is a finite sum `Σ ±aᵢ * bᵢ`, where `aᵢ` is a subposition of `x`, and `bᵢ` is a subposition of `y`. We create the auxiliary type `MulTy α β = Multiset (Bool × α × β)` of formal signed sums of products. We're then able to describe the product of two games `corec lα rα x` and `corec lβ rβ y` as the corecursor applied to two specific relations in `MulTy α β`.

I've proven basically nothing other than the characterizing theorem of multiplication. This PR is already long and convoluted enough as is!

- [x] depends on: #169
- [x] depends on: #172


CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -570,10 +574,43 @@ theorem add_dud (x : LGame) : x + dud = dud := by
 theorem dud_add (x : LGame) : dud + x = dud := by
   rw [add_comm, add_dud]
 
+theorem leftMoves_sum (m : Multiset LGame) : m.sum.leftMoves =
+    ⋃ y ∈ m, (· + (m.erase y).sum) '' y.leftMoves := by
+  induction m using Multiset.induction with
+  | empty => simp
+  | cons a m IH =>
+    simp only [leftMoves_add, IH, image_iUnion, image_image, Multiset.mem_cons, Multiset.sum_cons,
+      iUnion_iUnion_eq_or_left, Multiset.erase_cons_head]
```
tristan-f-r:
```suggestion
    simp_rw [Multiset.mem_cons, Multiset.sum_cons, iUnion_iUnion_eq_or_left,
      Multiset.erase_cons_head, leftMoves_add, IH, image_iUnion, image_image]
```

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -570,10 +574,43 @@ theorem add_dud (x : LGame) : x + dud = dud := by
...
+    rw [Multiset.erase_cons_tail_of_mem h]
+    simp [← add_assoc, add_comm]
+
+theorem rightMoves_sum (m : Multiset LGame) : m.sum.rightMoves =
+    ⋃ y ∈ m, (· + (m.erase y).sum) '' y.rightMoves := by
+  induction m using Multiset.induction with
+  | empty => simp
+  | cons a m IH =>
+    simp only [rightMoves_add, IH, image_iUnion, image_image, Multiset.mem_cons, Multiset.sum_cons,
+      iUnion_iUnion_eq_or_left, Multiset.erase_cons_head]
```
tristan-f-r:
```suggestion
    simp_rw [Multiset.mem_cons, Multiset.sum_cons, iUnion_iUnion_eq_or_left,
      Multiset.erase_cons_head, rightMoves_add, IH, image_iUnion, image_image]
```

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -612,4 +649,432 @@ theorem off_sub_off : off - off = dud := by
 theorem sub_dud (x : LGame) : x - dud = dud := by
   simp [sub_eq_add_neg]
 
+/-! ### Multiplication -/
+
+/-- Given two game graphs drawn on types `α` and `β`, the graph for the product can be drawn on the
+type `Multiset (Bool × α × β)`. Each term corresponds to a sum `Σ ±aᵢ * bᵢ`, where `aᵢ` and `bᵢ` are
+terms of `α` and `β` respectively. -/
+@[reducible]
+def MulTy (α β : Type*) :=
```
tristan-f-r:
```suggestion
abbrev MulTy (α β : Type*) :=
```

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -612,4 +649,432 @@ theorem off_sub_off : off - off = dud := by
 theorem sub_dud (x : LGame) : x - dud = dud := by
   simp [sub_eq_add_neg]
 
+/-! ### Multiplication -/
+
+/-- Given two game graphs drawn on types `α` and `β`, the graph for the product can be drawn on the
+type `Multiset (Bool × α × β)`. Each term corresponds to a sum `Σ ±aᵢ * bᵢ`, where `aᵢ` and `bᵢ` are
+terms of `α` and `β` respectively. -/
```
tristan-f-r:
```suggestion
terms of `α` and `β` respectively, and the attached bool represents each term's sign. -/
```

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -612,4 +649,432 @@ theorem off_sub_off : off - off = dud := by
...
+    mulOption b x y = {(b, y.1, x.2)} + {(b, x.1, y.2)} + {(!b, y.1, y.2)} :=
+  rfl
+
+variable (leftMovesα rightMovesα : α → Set α) (leftMovesβ rightMovesβ : β → Set β)
+
+/-- The set of pairs `α × β` used in `leftMovesSingle`. -/
+def leftMovesSet (x : Bool × α × β) : Set (α × β) :=
+  x.1.rec
+    (leftMovesα x.2.1 ×ˢ rightMovesβ x.2.2 ∪ rightMovesα x.2.1 ×ˢ leftMovesβ x.2.2)
+    (leftMovesα x.2.1 ×ˢ leftMovesβ x.2.2 ∪ rightMovesα x.2.1 ×ˢ rightMovesβ x.2.2)
```
tristan-f-r, line 716: This is more of a question: `rec` here is a red herring, and a match would be more appropriate here:

```suggestion
def leftMovesSet (x : Bool × α × β) : Set (α × β) := match x.1 with
  | true => (leftMovesα x.2.1 ×ˢ rightMovesβ x.2.2 ∪ rightMovesα x.2.1 ×ˢ leftMovesβ x.2.2)
  | false => (leftMovesα x.2.1 ×ˢ leftMovesβ x.2.2 ∪ rightMovesα x.2.1 ×ˢ rightMovesβ x.2.2)
```

however, lean doesn't immediately reduce this in the `rfl` closing proofs for `leftMoves_mul` and `rightMoves_mul` - is that why we prever not using a plain match statement here?

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -612,4 +649,432 @@ theorem off_sub_off : off - off = dud := by
...
+    (hlf : leftMovesα₂ ∘ f = Set.image f ∘ leftMovesα₁)
+    (hrf : rightMovesα₂ ∘ f = Set.image f ∘ rightMovesα₁)
+    (hlg : leftMovesβ₂ ∘ g = Set.image g ∘ leftMovesβ₁)
+    (hrg : rightMovesβ₂ ∘ g = Set.image g ∘ rightMovesβ₁) :
+    leftMovesSingle leftMovesα₂ rightMovesα₂ leftMovesβ₂ rightMovesβ₂ ∘ Prod.map id (Prod.map f g) =
+    image (map f g) ∘ leftMovesSingle leftMovesα₁ rightMovesα₁ leftMovesβ₁ rightMovesβ₁ := by
+  simp_rw [funext_iff, Function.comp_apply, leftMovesSingle, leftMovesSet] at *
+  rintro ⟨(_ | _), x⟩ <;> aesop
+
+-- TODO: find a faster proof using the previous theorem?
```
tristan-f-r, line 819: (thats if we prove that `rightMovesSingle` is an additive monoid homomorphism beforehand, which is a bit more annoying than this proof)

CombinatorialGames/Game/Loopy/Basic.lean
vihdzp: I'm surprised this doesn't work. I guess whatever internal representation `match` uses doesn't have defeqs quite as nice. 

I wouldn't call this a red herring, though. This is the recursor on the inductive type of booleans, which just so happens to correspond with a case-by-case construction!

CombinatorialGames/Game/Loopy/Basic.lean
vihdzp: My idea was to convert this into the `leftMoves` version by using negation, but that didn't quite work out.

There might be some other simple proof that avoids `aesop`, so I'll leave the TODO comment here.

CombinatorialGames/Game/Loopy/Basic.lean
tristan-f-r: This could be because I've only seen recursors when `motive` is wanted, but specifying the bool recursor seems to indicate that `motive` isn't just a constant, which is where the red herring comes from. It is fine here, though, since `match` doesn't work here though :+1:

PR #177, opened 2025-08-07
feat: begin natToFGame

tristan-f-r:
I'm marking this as a draft since there is no proof of `natToFGame`'s correctness. One quite nice thing about working with computable types is that you can quickly test functions like you would without a theorem prover:

```
-- true
#eval ((List.range 200).map (fun n ↦ FGameToNat (natToFGame n : FGame.{0}))) = List.range 200
```

### DIFF
--- a/CombinatorialGames/Game/Computability/Countable.lean
+++ b/CombinatorialGames/Game/Computability/Countable.lean
@@ -1,0 +1,146 @@
+/-
+Copyright (c) 2025 Tristan Figueroa-Reid. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Tristan Figueroa-Reid, Violeta Hernández
+-/
+import CombinatorialGames.Game.Computability.FGame
+import Mathlib.Data.Nat.Digits.Lemmas
+
+/-!
+# The computable bijection between ℕ and FGame
+
+Here, we provide a recursive, computable function constructed by Violeta Hernández from ℕ
+to FGame, derived from Ackermann’s Encoding, and, while it suffices to show that it's surjective
+to show its correctness for Plausible, we aim to prove that it is bijective.
+
+We then construct a `Plausible` instance on `FGame` using this function, allowing us to sample an ℕ
+from RNG and use that to construct an `FGame` for counterexamples.
+
+## Todo
+
+- Define plausible (This can be done right now, but `unsafe Repr` is weird. We can make a stable
+`Repr` by sorting over `FGameToNat`.)
+- Show that `natToFGame` is bijective with `FGameToNat`.
+-/
+
+/- ### For Mathlib -/
+
+instance List.decidableBExU {α : Type*} [DecidableEq α] (p : α → Prop) [DecidablePred p]
+    (l : List α) :
+    Decidable (∃! x, x ∈ l ∧ p x) :=
+  if h : (∃ x, x ∈ l ∧ p x) ∧ ∃ y ∈ l, ∀ x ∈ l, ¬x = y → ¬p x then .isTrue (by
+    obtain ⟨⟨x, hx⟩, ⟨z, hz⟩⟩ := h
+    refine ⟨x, hx, fun y hy ↦ ?_⟩
+    exact (not_imp_not.mp (hz.2 y hy.1) hy.2).trans (not_imp_not.mp (hz.2 x hx.1) hx.2).symm)
+  else .isFalse (by
+    unfold ExistsUnique
+    push_neg at h ⊢
+    intro x hx
+    have ⟨y, hy⟩ := h ⟨x, hx⟩ x hx.1
+    exact ⟨y, ⟨hy.1, hy.2.2⟩, hy.2.1⟩)
+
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Tristan Figueroa-Reid, Violeta Hernández
+-/
+import CombinatorialGames.Game.Computability.FGame
+import Mathlib.Data.Nat.Digits.Lemmas
+
+/-!
+# The computable bijection between ℕ and FGame
+
+Here, we provide a recursive, computable function constructed by Violeta Hernández from ℕ
```
vihdzp: I appreciate the citation, but given how similar this idea is to Ackermann's Encoding I don't think it's necessary.

CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
+/-
+Copyright (c) 2025 Tristan Figueroa-Reid. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Tristan Figueroa-Reid, Violeta Hernández
```
vihdzp:
```suggestion
Authors: Tristan Figueroa-Reid, Violeta Hernández Palacios
```

CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
...
+to FGame, derived from Ackermann’s Encoding, and, while it suffices to show that it's surjective
+to show its correctness for Plausible, we aim to prove that it is bijective.
+
+We then construct a `Plausible` instance on `FGame` using this function, allowing us to sample an ℕ
+from RNG and use that to construct an `FGame` for counterexamples.
+
+## Todo
+
+- Define plausible (This can be done right now, but `unsafe Repr` is weird. We can make a stable
+`Repr` by sorting over `FGameToNat`.)
```
vihdzp:
```suggestion
  `Repr` by sorting over `FGameToNat`.)
```

CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
...
+  inj' := fun _ ↦ by aesop
+
+theorem Placement.toNat_lt_four (p : Placement) : p.toNat < 4 := by unfold toNat; aesop
+
+/-- Places a child `FGame` into a parent `FGame` following the `Placement` rule. -/
+def Placement.place (parent : FGame) (child : FGame) : Placement → FGame
+| None => parent
+| Left => {parent.leftMoves ∪ {child} | parent.rightMoves}ꟳ
+| Right => {parent.leftMoves | parent.rightMoves ∪ {child}}ꟳ
+| Both => {parent.leftMoves ∪ {child} | parent.rightMoves ∪ {child}}ꟳ
```
vihdzp: You should use `Set.insert` for this.
```suggestion
| Left => {insert child (parent.leftMoves) | parent.rightMoves}ꟳ
| Right => {parent.leftMoves | insert child (parent.rightMoves)}ꟳ
| Both => {insert child (parent.leftMoves) | insert child (parent.rightMoves)}ꟳ
```

CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
...
+
+/-- Converts a natural number to a placement -/
+def Placement.fromNat {n : ℕ} (hn : n < 4) : Placement := match n with
+| 0 => None
+| 1 => Left
+| 2 => Right
+| 3 => Both
+
+/-- Converts a placement to a natural number -/
+def Placement.toNat : Placement ↪ Nat where
```
vihdzp: I feel like it could be nicer to instead define an equiv `Placement ≅ Fin 4`. Or even, could we work using `Fin 4` instead of `Placement`?

CombinatorialGames/Game/Computability/Countable.lean
```diff
@@ -0,0 +1,146 @@
...
+  obtain ⟨_, hy, rfl⟩ := hx
+  exact hy
+
+/-- The inverse of `natToFGame`. -/
+def FGameToNat (g : FGame) : ℕ :=
+  Nat.ofDigits 4 (((FGameToPlacements g).attach.image
+    fun ⟨x, _⟩ ↦ (FGameToNat x.1, x.2)).asList.map Placement.toNat)
+termination_by g
+decreasing_by exact .single (FGameToPlacements_mem_isOption (by assumption))
+
```
vihdzp, line 156: Can you add the `proof_wanted` for the result that's missing?

CombinatorialGames/Game/Computability/Countable.lean
tristan-f-r: We could work over `Fin 4`, but I don't think we lose much from having to deal with `Placement`. I do agree though that `Placement ≅ Fin 4` seems nicer here

PR #184, opened 2025-08-10
feat: more simp lemmas for `Real.toSurreal`

vihdzp:
I use these in a branch of mine.

### DIFF
--- a/CombinatorialGames/Surreal/Real.lean
+++ b/CombinatorialGames/Surreal/Real.lean
@@ -404,6 +404,11 @@ theorem toSurreal_max (x y : ℝ) : max x y = max (toSurreal x) (toSurreal y) :=
   have := le_total x y
   aesop
 
+@[simp]
+theorem toSurreal_min (x y : ℝ) : min x y = min (toSurreal x) (toSurreal y) := by
+  have := le_total x y
+  aesop
+
 @[simp, norm_cast]
 theorem toSurreal_abs (x : ℝ) : |x| = |toSurreal x| := by
   simp [abs]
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Surreal/Real.lean
```diff
@@ -388,12 +388,25 @@ theorem toSurreal_ratCast (q : ℚ) : toSurreal q = q := by
...
+  Surreal.mk_eq (toIGame_sub_equiv x y)
+
+@[simp]
+theorem toSurreal_max (x y : ℝ) : max x y = max (toSurreal x) (toSurreal y) := by
+  have := le_total x y
+  aesop
+
+@[simp, norm_cast]
+theorem toSurreal_abs (x : ℝ) : |x| = |toSurreal x| := by
+  simp [abs]
```
tristan-f-r: Am I missing something here? The last two simp directions go against `toSurreal_add` and `toSurreal_sub`.

```suggestion
@[simp]
theorem toSurreal_max (x y : ℝ) : max (toSurreal x) (toSurreal y) = max x y := by
  have := le_total x y
  aesop

@[simp, norm_cast]
theorem toSurreal_abs (x : ℝ) : |toSurreal x| = |x| := by
  simp [abs]
```

tristan-f-r (changes requested):
?

CombinatorialGames/Surreal/Real.lean
vihdzp: The lemma `toSurreal_max` actually reads
```
toSurreal (max x y) = max (toSurreal x) (toSurreal y)
```
So it does match `toSurreal_add` and all the others. I'm just removing the first `toSurreal` since it can be inferred. If it makes the lemma confusing to read I can add it back.

PR #178, opened 2025-08-08
feat: stopping time of loopy games

plp127:
- [x] Depends on: #279

PR #185, opened 2025-08-11
feat: lemmas on `WithTop`

vihdzp:
This should all be upstreamed to Mathlib eventually (with `NatOrdinal` replaced with `Ordinal`). These lemmas help golf the results on birthdays on cuts, and should be helpful with #178 as well.


CombinatorialGames/Mathlib/WithTop.lean
```diff
@@ -0,0 +1,72 @@
...
+  · rw [lt_top_iff_ne_top, ne_top_iff_exists] at h
+    obtain ⟨a, ha⟩ := h
+    rw [range_comp']
+    refine @small_image _ _ _ _ (bddAbove_iff_small.1 ⟨a, ?_⟩)
+    rintro _ ⟨i, rfl⟩
+    rw [← coe_le_coe, ha]
+    apply le_iSup _ i
+  · rw [range_comp'] at hg'
+    have : Small.{u} (range g) := small_of_image coe_injective.injOn
+    rw [iSup, range_comp']
```
tristan-f-r: (just a reordering to make it clear that `withTop_sSup_coe_lt` uses this instance)

```suggestion
    rw [iSup, range_comp']
    have : Small.{u} (range g) := small_of_image coe_injective.injOn
```

PR #188, opened 2025-08-11
feat: `ordinal_alias!` macro

vihdzp:
We collect all the boilerplate results which set up the type aliases `NatOrdinal` and `Nimber` into a single macro `ordinal_alias!`. Besides saving us lines of code, this helps ensure that the APIs for both aliases stay in sync.

This is my first foray into metaprogramming, so lmk if there's a better way to do what I did.

- [x] depends on: #186


tristan-f-r (approved):
I've also never done metaprogramming in lean before, though this looks like how I would do it in other languages.

I believe that @plp127 has done metaprogramming with lean, so it might be best for @plp127 to review this as well 👍

plp127 (approved):
I haven't caught any problems yet, so it's fine, probably?

PR #192, opened 2025-08-15
feat: w^w^w algebraically closed

DjangoPeeters:
This pull request tries to prove that w^w^w is algebraically closed.

### DIFF
--- a/CombinatorialGames/Nimber/WWW.lean
+++ b/CombinatorialGames/Nimber/WWW.lean
@@ -1,0 +1,130 @@
+/-
+Copyright (c) 2025 Django Peeters. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Django Peeters
+-/
+import Mathlib.Data.Nat.Prime.Defs
+import Mathlib.Data.Nat.Prime.Nth
+import Mathlib.Data.Nat.Nth
+import CombinatorialGames.Nimber.Simplicity
+import Mathlib.Fieldtheory.Minpoly.Field
+import Mathlib.Algebra.Field.Subfield.Defs
+import Mathlib.NumberTheory.PrimeCounting
+import Mathlib.SetTheory.Ordinal.CantorNormalForm
+
+/-!
+# w^(w^w) is algebraically closed
+
+This file aims to prove that `w ^ (w ^ w)` is an algebraically closed field.
+
+The proof follows [On Numbers And Games][conway2001] (pp. 59-61)
+and [On The Algebraic Closure Of Two][lenstra1977] (pp. 1-8).
+
+## Todo
+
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Nimber/WWW.lean
```diff
@@ -0,0 +1,130 @@
+/-
+Copyright (c) 2025 Django Peeters. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Django Peeters
+-/
+import Mathlib.Data.Nat.Prime.Defs
+import Mathlib.Data.Nat.Prime.Nth
+import Mathlib.Data.Nat.Nth
+import CombinatorialGames.Nimber.Simplicity
+import Mathlib.Fieldtheory.Minpoly.Field
```
tristan-f-r:
```suggestion
import Mathlib.FieldTheory.Minpoly.Field
```

tristan-f-r (commented):
(failed import)

CombinatorialGames/Nimber/WWW.lean
DjangoPeeters: Huh, weird. Both ways work on my laptop. Any idea why?

tristan-f-r: Perhaps its because Windows filepaths are case insensitive?

CombinatorialGames/Nimber/WWW.lean
```diff
@@ -0,0 +1,138 @@
...
+  (Nimber.kappa (nth Nat.Prime k)) ^ (nth Nat.Prime k) = Nimber.alpha (nth Nat.Prime k) := by
+  sorry
+
+protected theorem kappa_pow_decrease (k n : ℕ) {h : n ≠ 0} :
+  (Nimber.kappa ((nth Nat.Prime k) ^ (n + 1))) ^ (nth Nat.Prime k)
+  = (Nimber.kappa ((nth Nat.Prime k) ^ n)) := by
+  sorry
+
+private def tau' : Nimber :=
+  sInf {x | IsAlgClosed x}
```
vihdzp: This is actually equal to `0` because of how we treat those edge cases. Oops.

To prove that `Nimber` is algebraically closed, I'll introduce a definition `algClosure : Nimber → Nimber`. You can then state `algClosure (∗2) = tau`.

DjangoPeeters: Are we allowing 0 (and sometimes 1) to be a group/ring/field/algebraically closed because it's less overhead?

vihdzp: Really it's a matter of "the important theorems don't require these assumptions, so we don't bother adding them".

PR #193, opened 2025-08-16
feat: `AddCommMonoidWithOne IGame` and `MulZeroOneClass IGame`

astrainfinita:


CombinatorialGames/Game/IGame.lean
```diff
@@ -1081,6 +1081,8 @@ instance : MulOneClass IGame where
   one_mul := one_mul'
   mul_one x := mul_comm' .. ▸ one_mul' x
 
+instance : MulZeroOneClass IGame where
```
vihdzp: I think you could simply turn the `MulOneClass` instance into a `MulZeroOneClass` instance.

PR #196, opened 2025-08-17
chore: rename and split `Nimber.Simplicity`

vihdzp:
We create three new files:

- `Nimber.SimplestExtension.Basic`, holding the proof of the first three simplest extension theorems.
- `Nimber.SimplestExtension.Polynomial`, which will hold a bunch of auxiliary definitions for polynomials (to be PR'd later).
- `Nimber.SimplestExtension.Algebraic`, which will contain the proof of the last part of the simplest extension theorem, as well as the proof that nimbers are algebraically closed.

There's no further mathematics done in this PR.


alreadydone:
Why don't you rename the folder SimplestExtension?

vihdzp:
Oh, that explains the confusion! We actually do have that, see e.g. [`IGame.Fits.equiv_of_forall_not_fits`](https://vihdzp.github.io/combinatorial-games/CombinatorialGames/Surreal/Basic.html#IGame.Fits.equiv_of_forall_not_fits).

alreadydone:
The title should become something like "rename and split `Nimber.Simplicity`"

DjangoPeeters (approved):
These changes look good to me.

PR #200, opened 2025-08-18
feat: `IsField.embed`

vihdzp:
We introduce `IsField.embed`, which reinterprets a polynomial as existing in the subfield for an element `IsField x`.

We could have introduced this as `IsRing.embed` instead, but that has the trouble that Lean can't infer a field instance for `h.toSubring` even when `h : IsField x`, and some later results will require a field.

### DIFF
--- a/CombinatorialGames/Nimber/SimplestExtension/Polynomial.lean
+++ b/CombinatorialGames/Nimber/SimplestExtension/Polynomial.lean
@@ -38,38 +38,6 @@ theorem natDegree_eq_zero_iff : p.natDegree = 0 ↔ p = 0 ∨ p.degree = 0 := by
   rw [p.natDegree_eq_zero_iff_degree_le_zero, le_iff_lt_or_eq, ← WithBot.coe_zero, ← bot_eq_zero',
     WithBot.lt_coe_bot, p.degree_eq_bot]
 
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Nimber/SimplestExtension/Polynomial.lean
```diff
@@ -31,3 +38,96 @@ theorem natDegree_eq_zero_iff : p.natDegree = 0 ↔ p = 0 ∨ p.degree = 0 := by
...
+theorem IsRing.coeff_prod_lt {x : Nimber} {ι} {f : ι → Nimber[X]} {s : Finset ι} (h : IsRing x)
+    (hx₁ : 1 < x) (hs : ∀ y ∈ s, ∀ k, (f y).coeff k < x) : ∀ k, (s.prod f).coeff k < x := by
+  classical
+  induction s using Finset.induction with
+  | empty =>
+    have := zero_lt_one.trans hx₁
+    aesop
+  | insert y s hy IH =>
+    rw [Finset.prod_insert hy]
+    apply h.coeff_mul_lt <;> aesop
```
tristan-f-r:
```suggestion
    apply h.coeff_mul_lt <;> simp_all
```

tristan-f-r (approved):
(Thanks for explaining the reasoning for choosing `IsField` over `IsRing`!)

PR #204, opened 2025-08-20
feat: nim operations on naturals

vihdzp:
We define a type alias for the naturals endowed with nimber operations. We define addition and multiplication (though only the former is currently proven correct).

### DIFF
--- a/CombinatorialGames/Nimber/Nat.lean
+++ b/CombinatorialGames/Nimber/Nat.lean
@@ -1,5 +1,5 @@
 /-
-Copyright (c) 2024 Violeta Hernández Palacios. All rights reserved.
+Copyright (c) 2025 Violeta Hernández Palacios. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Violeta Hernández Palacios
 -/
--- a/CombinatorialGames/Nimber/Nat.lean
+++ b/CombinatorialGames/Nimber/Nat.lean
@@ -1,0 +1,58 @@
+import CombinatorialGames.Nimber.Basic
+
+alias! /-- The natural numbers, endowed with nim operations. -/ NatNimber Nat
+
+namespace NatNimber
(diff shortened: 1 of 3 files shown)


CombinatorialGames/Nimber/Nat.lean
```diff
@@ -0,0 +1,94 @@
...
+instance : Add NatNimber where
+  add m n := of (m.val ^^^ n.val)
+
+instance : Sub NatNimber where
+  sub m n := m + n
+
+@[simp]
+theorem toNimber_add (x y : NatNimber) : toNimber (x + y) = .of x.val + .of y.val :=
+  (Nimber.add_nat ..).symm
+
```
tristan-f-r, line 54: This is obviously true, but it's nice to see a proof attached to the `Sub` and `Neg` instances.

```suggestion

@[simp]
theorem toNimber_sub (x y : NatNimber) : toNimber (x - y) = .of x.val - .of y.val :=
  toNimber_add ..
  
@[simp]
theorem toNimber_neg (x : NatNimber) : toNimber (-x) = .of x.val :=
  Nimber.neg_eq ..

```

tristan-f-r (approved):

CombinatorialGames/Nimber/Nat.lean
vihdzp: I think it'd be better to have simp lemmas `x - y = x + y` and `-x = x` instead. (These are implied by them)

tristan-f-r: Sounds good :+1:

tristan-f-r:
```suggestion

@[simp] theorem sub_eq (x y : NatNimber) : x - y = x + y := rfl
@[simp] theorem neg_eq (x : NatNimber) : -x = x := rfl

```

PR #207, opened 2025-08-21
chore: remove redundant theorems on game equivalence 

vihdzp:
There's two things that irk me about `equiv_of_forall_lf` and `equiv_of_exists_le`:
1. These are actually mathematical equivalences, and I think they should have been stated as such. In the context of a longer proof, I would not feel comfortable applying these theorems without knowing they were iffs - what if I end up with a goal that's false?
2. They aren't really worth stating as iffs either, since they're straightforward to show by just unfolding definitions (admittedly you have to be slightly careful since `simp_rw [le_iff_forall_lf]` loops, but as the new proof of `equiv_of_exists` shows this isn't a huge problem).

Asking for @plp127's approval since he's the one who added these in the first place.

### DIFF
--- a/CombinatorialGames/Game/IGame.lean
+++ b/CombinatorialGames/Game/IGame.lean
@@ -453,26 +453,15 @@ def delabFuzzy : Delab := do
     annotateGoToSyntaxDef stx
   catch _ => failure -- fail over to the default delaborator
 
-theorem equiv_of_forall_lf {x y : IGame}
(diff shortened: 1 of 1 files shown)


plp127:
I was gonna use those!

vihdzp:
Well, my question is whether you really need them. Isn't it easier to just use `constructor` to split `AntisymmRel` and then unfold ≤ as needed?

PR #208, opened 2025-08-22
feat: `OfSets` notation class

astrainfinita:


vihdzp (approved):
This is awesome! It's great having a unified notation (and it's also great `Surreal.ofSets` doesn't try `aesop` willy-nilly). I'll just let @tristan-f-r take a look at this before merging.

CombinatorialGames/Game/Functor.lean
```diff
@@ -119,6 +119,64 @@ instance : CommGroup Player where
...
+      `(!{$s | $t})
+    else
+      let st ← delab
+      `(!{$st})
+
+theorem ofSets_eq_ofSets_cases {α} {Valid : (Player → Set α) → Prop} [OfSets α Valid]
+    (st : Player → Set α) (h : Valid st) [Small (st left)] [Small (st right)] :
+    !{st} = !{st left | st right}'(by convert h; aesop) := by
+  congr; ext1 p; cases p <;> rfl
+
```
vihdzp, line 183: This file now seems to consist of two things:
- Auxiliary types/notation to deal with general games
- QPF shenanigans

Can you add some `/-! ### Sectioning -/` to demarcate these two parts?

tristan-f-r (approved):
I also really like this! I would like to see this in `FGame` as well, though it's understandable not to add it there since that file is quite stale now.

CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -784,14 +777,18 @@ theorem map_mulOption (b x y) :
...
 theorem movesSingle_comp_prodMap
     (hf : ∀ p, movesα₂ p ∘ f = Set.image f ∘ movesα₁ p)
     (hg : ∀ p, movesβ₂ p ∘ g = Set.image g ∘ movesβ₁ p) :
     movesSingle p movesα₂ movesβ₂ ∘ Prod.map id (Prod.map f g) =
     image (map f g) ∘ movesSingle p movesα₁ movesβ₁ := by
   simp_rw [funext_iff, Function.comp_apply, movesSingle, movesSet] at *
-  rintro ⟨(_ | _), x⟩ <;> aesop
+  rintro ⟨p', x⟩
+  ext
+  simp only [Prod.map_apply, id_eq, Prod.map_fst, Prod.map_snd, mem_image, Prod.exists, hf, hg]
```
tristan-f-r:
```suggestion
  simp_rw [Prod.map_apply, id_eq, Prod.map_fst, Prod.map_snd, mem_image, Prod.exists, hf, hg]
```

CombinatorialGames/Game/Functor.lean
```diff
@@ -119,6 +121,68 @@ instance : CommGroup Player where
...
+class OfSets (α : Type (u + 1)) (Valid : outParam ((Player → Set α) → Prop)) where
+  /-- Construct a combinatorial game from its left and right sets. -/
+  ofSets (st : Player → Set α) (h : Valid st) [Small.{u} (st left)] [Small.{u} (st right)] : α
+export OfSets (ofSets)
+
+@[inherit_doc OfSets.ofSets]
+macro "!{" st:term "}'" h:term:max : term => `(OfSets.ofSets $st $h)
+
+@[inherit_doc OfSets.ofSets]
+macro "!{" s:term " | " t:term "}'" h:term:max : term => `(!{Player.cases $s $t}'$h)
```
alreadydone, line 141: I think I like {}ᴳ better than !{} ...

astrainfinita: I also feel that the current notations are not good enough. However, we plan to use the notations `^L` and `^R` to denote `moves left` and `moves right`, so the suffix `^G` is probably not appropriate.

vihdzp: I agree, the superscripts aren't really compatible with that new notation.

PR #211, opened 2025-08-22
feat: symmetrize `nim` API

vihdzp:
The size of this file is rapidly tending towards zero! I've added a `proof_wanted` which should help with that.

This is all possible thanks to @astrainfinita's refactor.

### DIFF
--- a/CombinatorialGames/Game/Functor.lean
+++ b/CombinatorialGames/Game/Functor.lean
@@ -46,7 +46,7 @@ inductive Player where
   | left  : Player
   /-- The Right player. -/
   | right : Player
-deriving DecidableEq
+deriving DecidableEq, Inhabited
 
 namespace Player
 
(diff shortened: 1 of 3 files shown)


CombinatorialGames/Game/Player.lean
```diff
@@ -26,7 +26,7 @@ inductive Player where
   | left  : Player
   /-- The Right player. -/
   | right : Player
-deriving DecidableEq
+deriving DecidableEq, Inhabited
```
tristan-f-r, line 29: This doesn't seem like a good instance. It does have the benefit of abstracting the player choice (which you've demonstrated in `nim_injective`), but, afaik, `Inhabited` holds the _designated_ default type, which seems quite problematic for some other uses of `Player` other than your example.

If this is a normal usage of `Inhabited`, then I'm fine with it. Otherwise, perhaps a `Nonempty Player` instance is fine? It looks a little worse than Inhabited, but it gives the same encapsulation effect.

vihdzp: There is a designated default term here: it's the one from the first constructor, `left`. Considering that `left` is the multiplicative identity, that seems like a reasonable choice to me.

`Nonempty` is equivalent to `Inhabited` under choice, but it does carry a different connotation, being "there exists some term, even if I can't exhibit even a single explicit example". This clearly isn't the case here.

tristan-f-r: I was under the (false) impression that, while you've provided a multiplication structure on `Player`, that was merely for the convenience of stating definitions, and not to actually canonicalize `left` as the preferred default instance. Since that's the case, I'm fine with this change.

PR #212, opened 2025-08-23
chore: prefer `!{fun _ ↦ s}` over `!{s | s}`

vihdzp:
The former allows `x.moves p` to simplify directly to `s`, while the latter doesn't.

### DIFF
--- a/CombinatorialGames/Game/Specific/Nim.lean
+++ b/CombinatorialGames/Game/Specific/Nim.lean
@@ -46,7 +46,7 @@ noncomputable def nim : Nimber.{u} → IGame.{u} :=
   ConcreteGame.nim.toIGame
 
 theorem nim_def (o : Nimber) : nim o = !{fun _ ↦ nim '' Iio o} :=
-  ConcreteGame.toIGame_def' o
+  ConcreteGame.toIGame_def' ..
 
 @[simp]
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Game/Loopy/Basic.lean
```diff
@@ -320,13 +320,16 @@ theorem rightMoves_ofSets (l r : Set LGame) [Small.{u} l] [Small.{u} r] :
...
-instance : Zero LGame := ⟨!{∅ | ∅}⟩
+instance : Zero LGame := ⟨!{fun _ ↦ ∅}⟩
 
-theorem zero_def : (0 : LGame) = !{∅ | ∅} := rfl
+theorem zero_def : (0 : LGame) = !{∅ | ∅} := ofSets_eq_ofSets_cases ..
 
 @[simp] theorem leftMoves_zero : leftMoves 0 = ∅ := leftMoves_ofSets ..
 @[simp] theorem rightMoves_zero : rightMoves 0 = ∅ := rightMoves_ofSets ..
 
+-- TODO: remove the former?
```
tristan-f-r, line 330: (This is about `left`/`rightMoves_zero`, right?)

vihdzp: Yep. I believe that @astrainfinita's planned refactor of using notation for `moves left` and `moves right` will remove those lemmas as a corollary.

PR #225, opened 2025-08-28
chore: symmetrize `Short` API further

vihdzp:
This is essentially the same reasoning as #224.


tristan-f-r (approved):
(sorry for the wait!)

PR #231, opened 2025-09-06
feat: evaluate a nimber polynomial using ordinal arithmetic

vihdzp:
We define `Polynomial.oeval`, which evaluates a polynomial using the arithmetic of the Cantor normal form.

The proofs and definitions here all kind of suck. I'm currently working on adding a function `Ordinal.CNF.eval : (Ordinal →₀ Ordinal) → Ordinal` to Mathlib; this would generalize this construction and would off-load the burden of proving all of these results about it. That's probably a month or two away, though, and I think we'd all rather make use of the simplest extension theorem now rather than later.


DjangoPeeters (approved):
I don't see anything wrong with this. The theorems look understandable to me.

CombinatorialGames/Nimber/SimplestExtension/Polynomial.lean
```diff
@@ -30,6 +31,29 @@ open Order Polynomial
...
+      apply add_left_mono
+      rw [← hm]
+      exact bot_le
+    · specialize ih h₁
+      simp only [List.sum_cons]
+      apply ih.trans
+      conv_lhs => rw [← zero_add ys.sum]
+      apply add_right_mono
+      rw [← hm]
+      exact bot_le
```
tristan-f-r: Move the `mem_cons` and `sum_cons` calls to the top

```suggestion
    rw [List.mem_cons] at h₁
    rw [List.sum_cons]
    rcases h₁ with rfl | h₁
    · conv_lhs => rw [← add_zero x]
      apply add_left_mono
      rw [← hm]
      exact bot_le
    · specialize ih h₁
      apply ih.trans
      conv_lhs => rw [← zero_add ys.sum]
      apply add_right_mono
      rw [← hm]
      exact bot_le
```

tristan-f-r (approved):
I can't really golf much here, and the definitions seem correct.

PR #232, opened 2025-09-06
feat: make definitions computable

vihdzp:
A recent Mathlib PR made [`Shrink`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Small/Defs.html#Shrink) computable, which through a domino effect now means that a lot of our basic definitions on games are computable too.

This doesn't really do much for us. But at least we don't have to write `noncomputable section` literally everywhere!

### DIFF
--- a/CombinatorialGames/Game/Special.lean
+++ b/CombinatorialGames/Game/Special.lean
@@ -12,13 +12,12 @@ import CombinatorialGames.Tactic.GameCmp
 This file defines some simple yet notable combinatorial games:
 
 * `⋆ = {0 | 0}`
+* `¼ = {0 | 1}`
 * `↑ = {0 | ⋆}`
-* `↓ = {⋆ | 0}`.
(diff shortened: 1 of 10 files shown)


CombinatorialGames/Game/Special.lean
```diff
@@ -12,14 +12,13 @@ import CombinatorialGames.Tactic.GameCmp
 This file defines some simple yet notable combinatorial games:
 
 * `⋆ = {0 | 0}`
+* `¼ = {0 | 1}`
```
tristan-f-r:
```suggestion
```

tristan-f-r (approved):

CombinatorialGames/Game/Special.lean
vihdzp: whoops, this snuck in from another PR.

PR #236, opened 2025-09-09
feat: stoppers

vihdzp:
Should `Stopper` and `StopperFor` be typeclasses? I'm leaning towards no, since they're much less compositional than other typeclasses like `Numeric` or `Impartial`, but I'd like to hear opinions on this.


tristan-f-r (approved):
Not having `Stopper`/`StopperFor` be typeclasses make sense. It's easier to refactor a type to being a class than vice versa, and, as you said, it seems that the only good candidate for a stopper instance is `Stopper.neg`, which isn't much.

PR #277, opened 2025-12-19
feat: `Fits x y → birthday y ≤ birthday x`

vihdzp:
Split off from #263. This seems golfable. Also, is there some way to prove that `x` is the simplest game that fits within itself, even among non-numeric games?

- [x] depends on: #271


plp127:
> This seems golfable.

Do we have the theorem that if `x < y` are surreals with the same birthday, then `!{{x} | {y}}` has a smaller birthday? I think this should be sufficient.

> Also, is there some way to prove that `x` is the simplest game that fits within itself, even among non-numeric games?

Does it hold that if `x` and `y` are cuts with `h : x < y` are cuts of the same birthday then `birthday (simplestBtwn h) + 1` is strictly smaller? Of course making an exception for the case of `x = ⊥ ∧ y = ⊤`.

plp127:
This should follow trivially from the theorem `Fits.exists_wsubposition_equiv` proved in #435.

CombinatorialGames/Surreal/Birthday/Basic.lean
```diff
@@ -90,6 +90,24 @@ theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] : birthday ofNat(n) = n :=
 theorem birthday_one : birthday 1 = 1 := by
   simpa using birthday_natCast 1
 
+theorem _root_.IGame.Fits.birthday_le {x y : IGame} [Numeric x] [Numeric y] (h : Fits x y) :
```
plp127: Can you remove the `[Numeric y]` assumption here?

CombinatorialGames/Surreal/Birthday/Basic.lean
```diff
@@ -90,6 +90,24 @@ theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] : birthday ofNat(n) = n :=
...
+    birthday (.mk y) ≤ birthday (.mk x) := by
+  obtain ⟨x', _, hx', hxb⟩ := Surreal.birthday_eq_iGameBirthday (.mk x)
+  obtain ⟨z, hzx, hzy⟩ := (h.congr (mk_eq_mk.1 hx'.symm)).exists_wsubposition_equiv
+  have := Numeric.wsubposition hzx
+  rw [← Surreal.mk_eq hzy, ← hxb]
+  exact (birthday_mk_le z).trans <| birthday_le_of_wsubposition hzx
+
+theorem _root_.IGame.Fits.birthday_lt {x y : IGame} [Numeric x] [Numeric y]
```
plp127: Same here, `y` shouldn't need to be numeric.

CombinatorialGames/Surreal/Birthday/Basic.lean
```diff
@@ -90,6 +90,24 @@ theorem birthday_ofNat (n : ℕ) [n.AtLeastTwo] : birthday ofNat(n) = n :=
 theorem birthday_one : birthday 1 = 1 := by
   simpa using birthday_natCast 1
 
+theorem _root_.IGame.Fits.birthday_le {x y : IGame} [Numeric x] [Numeric y] (h : Fits x y) :
+    birthday (.mk y) ≤ birthday (.mk x) := by
+  obtain ⟨x', _, hx', hxb⟩ := Surreal.birthday_eq_iGameBirthday (.mk x)
+  obtain ⟨z, hzx, hzy⟩ := (h.congr (mk_eq_mk.1 hx'.symm)).exists_wsubposition_equiv
+  have := Numeric.wsubposition hzx
```
plp127: This something that the `numeric` tactic should be doing, can we fix that.

plp127 (approved):

CombinatorialGames/Surreal/Birthday/Basic.lean
vihdzp: We can, and it's not difficult, but actually this hypothesis is unused! I'll do it in another PR.

vihdzp: #445

PR #292, opened 2026-01-04
feat: indexing the support of `SurrealHahnSeries` by ordinals

vihdzp:
A lot of the proofs here are kind of horrible. On a subsequent PR I will introduce `SurrealHahnSeries.ofSeq` for creating a Hahn series out of a decreasing ordinal-indexed sequence of exponents and coefficients; this turns out to be a much more convenient representation for proving results about length.

We opt to define `SurrealHahnSeries.length` as an `Ordinal` rather than a `NatOrdinal`.  The reasoning being that the former is a more natural type to describe order types, and the conversion is basically trivial anyways.


vihdzp:
I have an idea to redesign this file. I think I'll implement that instead.

wwylele (approved):
I am not well-versed in ordinals and set theory, so I can only say this looks good based my surface-level intuition.

PR #298, opened 2026-01-11
feat: `simplestBtwn x y` is in fact the simplest surreal between `x` and `y`

vihdzp:
We also change `simplestBtwn` to take two surreals `x y` as arguments, rather than a proof `x < y`.

- [x] depends on: #277
- [x] depends on: #297


CombinatorialGames/Surreal/Cut.lean
```diff
@@ -474,43 +474,67 @@ theorem Fits.rightSurreal_le {x : Surreal} {y z : Cut} (h : Fits x y z) : rightS
...
   rw [Fits, ← mem_compl_iff, compl_inter, compl_left, compl_right]
 
-/-- The simplest surreal number (in terms of birthday) that fits between two cuts. -/
-noncomputable def simplestBtwn {x y : Cut} (h : x < y) : Surreal :=
-  Classical.choose <|
-    exists_minimalFor_of_wellFoundedLT _ birthday (lt_iff_nonempty_inter.1 h)
+/-- The simplest surreal number (in terms of birthday) that fits between two cuts.
+
+If none exists, we set a junk value of `0`. -/
+noncomputable def simplestBtwn (x y : Cut) : Surreal :=
```
plp127, line 555: ~~This is strictly less general, right?~~ misread

PR #306, opened 2026-01-13
feat: short stops are numeric

vihdzp:
This condition is relevant to e.g. the number translation theorem. The terminology is from Siegel.


CombinatorialGames/Surreal/Cut.lean
```diff
@@ -527,5 +625,207 @@ theorem supLeft_lt_infRight_of_equiv_numeric {x y : IGame} [y.Numeric] (h : x 
...
+theorem rightGame_switch {x : IGame} (h : 0 ≤ x) [Numeric x] :
+    rightGame (.mk (±x)) = -rightSurreal (.mk x) := by
+  rw [rightGame_eq_infRight_of_le (infRight_switch_le_supLeft_switch h)]
+  simp
+
+/-! ### Numeric cuts -/
+
+/-- We say that a cut is numeric when it's either a left or right cut from a surreal. -/
+@[mk_iff numeric_def']
+protected class inductive Numeric : Cut → Prop where
```
plp127, line 609: Where did this name come from?

vihdzp: Siegel, section VIII, exercise 1.7. Page 410. "A section (X, Y) is numeric if it has the form L(x) or R(x), for some surreal
number x." A section here being an alternate name for a surreal cut.

In fact, this PR proves part (b) from that exercise :)

CombinatorialGames/Surreal/Cut.lean
```diff
@@ -407,6 +475,10 @@ theorem right_supLeft (x : IGame) :
     (supLeft x).right = ⋂ i ∈ xᴸ, {y | .mk i ⧏ y.toGame} := by
   simp [supLeft]
 
+theorem supLeft_mem_of_short {x : IGame} [Short x] (hx : xᴸ.Nonempty) :
+    ∃ y ∈ xᴸ, rightGame (.mk y) = supLeft x :=
```
plp127, line 478: In a future PR, might we merge `leftGame` and `rightGame` with a `Player` argument?

plp127 (approved):

CombinatorialGames/Surreal/Cut.lean
vihdzp: I half-remember there already being a PR for this? And if there isn't then yeah! we should do this.

PR #317, opened 2026-01-26
feat: omnific integers

vihdzp:
I'm going to reuse the `round` function to prove some stuff about Hahn series, which is why I wanted it to have a name.


CombinatorialGames/Surreal/Omnific/Basic.lean
```diff
@@ -0,0 +1,211 @@
...
+noncomputable section
+
+/-! ### For Mathlib -/
+
+namespace Set
+variable {α β : Type*}
+
+@[simp]
+theorem range_singleton (x : α) (f : ({x} : Set α) → β) : range f = {f ⟨x, mem_singleton x⟩} := by
+  aesop
```
plp127: This id `Set.range_unique`

vihdzp: `range_unique` isn't tagged as `simp` for what I presume are performance reasons. So I think it makes sense to keep `range_singleton` as a `simp` lemma, just golfing its proof.

CombinatorialGames/Game/Classes.lean
```diff
@@ -407,21 +407,23 @@ decreasing_by igame_wf
 theorem neg_iff {x : IGame} : Numeric (-x) ↔ Numeric x :=
   ⟨fun _ ↦ by simpa using Numeric.neg (-x), fun _ ↦ Numeric.neg x⟩
 
+@[aesop forward safe]
```
plp127: Is this safe (no pun intended) ? It seems to me like this could start creating an unbounded amount of `Numeric (x + (x + x) + (x + (x + x) + x + x) + (x + x + x))`-kind of hypotheses to populate the context with.

vihdzp: I'll admit I don't actually know what this attribute does, only that it helped `aesop` prove that `{x - y | x + y}` is numeric for `x`, `y` numeric with `y > 0`. Should I tag it as unsafe instead?

vihdzp: Well, I made it unsafe, and all seems good.

plp127: I'm still worried about `forward unsafe`, I don't think we should be tagging things if we don't know what the tags do.

vihdzp: Ah, it was actually pretty easy to remove these tags by just using `simp` instead.

CombinatorialGames/Surreal/Omnific/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+  generalize_proofs at hx hy
+  simp only [ofSets_mul_ofSets, mk_ofSets, mulOption, singleton_prod_singleton,
+    union_singleton, image_insert_eq, image_singleton, range_singleton, range_insert]
+  congr <;> dsimp <;> rw [hx, hy] <;> grind
+
+/-! ### Omnific integers -/
+
+/-- An omnific integer is one such that `x = {x - 1 | x + 1}`. These are an analog of the integers
+to the surreals. -/
+def IsOmnific (x : Surreal) : Prop :=
```
plp127, line 114: Another option could be to make this private and only make `Omnific : Subring Surreal` public, then you would instead spell `IsOmfinic x` as `x ∈ Omnific`.

CombinatorialGames/Surreal/Omnific/Basic.lean
```diff
@@ -0,0 +1,196 @@
...
+    rw [IsOmnific, round_of_zero_mem] at h
+    · exact hx.ne h
+    · grind
+
+theorem IsOmnific.lt_one_iff_nonpos {x : Surreal} (h : IsOmnific x) : x < 1 ↔ x ≤ 0 := by
+  rw [← not_iff_not]
+  simpa using h.one_le_iff_pos
+
+/-- The subring of `IsOmnific` surreal numbers. -/
+def Omnific : Subring Surreal where
```
plp127: I guess this technically violates the naming convention since `Subring Surreal` isn't `Type _`? I think for similar declarations in mathlib they decided to go with lowercasing the name, for example `MeasureTheory.Lp.simpleFunc` or `unitInterval`.

plp127 (approved):

CombinatorialGames/Surreal/Omnific/Basic.lean
vihdzp: Having `IsOmnific` is nice since it allows for not notation. I imagine we'd only be using `omnific` to state results about the entire structure, i.e. stuff like `DecompositionMonoid omnific`.

CombinatorialGames/Surreal/Omnific/Basic.lean
vihdzp: I've made it lowercase.

PR #327, opened 2026-01-30
feat: group/ring/field closures of nimbers

vihdzp:


CombinatorialGames/Nimber/SimplestExtension/Closure.lean
```diff
@@ -0,0 +1,326 @@
...
+
+theorem addSubgroupClosure_Iio_eq_Iio (x : Nimber) :
+    ∃ y, (closure (Iio x) : Set Nimber) = Iio y := by
+  refine (isLowerSet_addSubgroupClosure (isLowerSet_Iio x)).eq_univ_or_Iio.resolve_left fun hx ↦ ?_
+  have H : Small (_ : Set _) := small_addSubgroupClosure (Iio x)
+  rw [hx, small_univ_iff] at H
+  cases not_small_nimber H
+
+/-- Returns the smallest `IsGroup` that's at least `x`. -/
+def groupClosure (x : Nimber) : Nimber := Classical.choose (addSubgroupClosure_Iio_eq_Iio x)
```
plp127, line 130: Is it easier at all if you use `if IsGroup x then x else of (2 ^ (log 2 (val x) + 1))`?

vihdzp: That just feels needlessly tedious. It's nicer to work with the abstract "is a subgroup" property than it is with ordinal powers of two.

plp127 (approved):

PR #399, opened 2026-03-26
feat: simplicity order

vihdzp:


CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -0,0 +1,322 @@
...
+/-- The identity function between `Simplicity` and `SignExpansion`. -/
+def val : Simplicity ≃ SignExpansion := Equiv.refl _
+
+@[simp, grind =] theorem of_val (x) : of (val x) = x := rfl
+@[simp, grind =] theorem val_of (x) : val (of x) = x := rfl
+
+instance : Bot Simplicity := ⟨of 0⟩
+
+@[simp, grind =] theorem of_zero : of 0 = ⊥ := rfl
+@[simp, grind =] theorem val_zero : val ⊥ = 0 := rfl
```
plp127:
```suggestion
@[simp, grind =] theorem val_bot : val ⊥ = 0 := rfl
```

CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -0,0 +1,322 @@
...
+@[simp, grind =] theorem coe_bot : ⇑(⊥ : Simplicity) = 0 := rfl
+theorem bot_apply (o) : (⊥ : Simplicity) o = 0 := rfl
+
+@[ext]
+protected theorem ext {x y : Simplicity} (hxy : ∀ o, x o = y o) : x = y :=
+  DFunLike.coe_injective (funext hxy)
+
+/-- A recursor for `Simplicity`. Use as `cases x`. -/
+@[elab_as_elim, cases_eliminator, induction_eliminator]
+protected def ind {motive : Simplicity → Sort*} (of : ∀ a, motive (of a)) (a) : motive a :=
```
plp127, line 96: Since this is a `def`, should we have a lemma `ind of (Simplicity.of a) = of a`

CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -0,0 +1,322 @@
...
+  isLUB_sSup s := by
+    change IsLUB s (dite ..)
+    split_ifs with h hs
+    · simp [isLUB_iff_le_iff, upperBounds, WithTop.forall, h]
+    · exact WithTop.isLUB_of_coe_preimage h (isLUB_sSup_of_bddAbove hs)
+    · simpa [isLUB_iff_le_iff, upperBounds, WithTop.forall, h, bddAbove_def] using hs
+
+instance : CompleteLattice (WithTop Simplicity) where
+  __ := WithTop.instBoundedOrder
+  __ := instCompleteSemilatticeInfWithTop
```
plp127, line 319: can you name this instance since it's being referred to by name

CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -0,0 +1,322 @@
...
+  __ := completeLatticeOfCompleteSemilatticeSup _
+
+end Simplicity
+
+/-! ### Ordinals -/
+
+namespace NatOrdinal
+variable {a b : NatOrdinal}
+
+theorem of_toSignExpansion_strictMono : StrictMono fun o : NatOrdinal ↦ Simplicity.of o := by
```
plp127, line 329: is there anything here that indicates `Simplicity`?

CombinatorialGames/SignExpansion/Simplicity.lean
vihdzp: I'm not sure what a more appropriate name than `Simplicity.instCompleteSemilatticeInfWithTop` would be.

CombinatorialGames/SignExpansion/Simplicity.lean
vihdzp: `of` here refers to `Simplicity.of`. Any better naming ideas for this theorem?

CombinatorialGames/SignExpansion/Simplicity.lean
plp127: Then, you can write `instance instCompleteSemilatticeInfWithTop : CompleteSemilatticeInf (WithTop Simplicity) where` on line 205, the point is to have a name, so that you're not relying on an automatically generated one

CombinatorialGames/SignExpansion/Simplicity.lean
plp127: lots of things have an `of` function. perhaps you could say `simplicityOf` instead? though that isn't much better...

vihdzp: To be fair, there is only one `of` function that can precompose with `toSignExpansion`.
(Can we not bikeshed too much on the name of this one theorem? We can change it later if need be.)

plp127 (approved):

PR #416, opened 2026-05-27
chore: update to `v4.30.0`

vihdzp:


CombinatorialGames/Mathlib/EtaSet.lean
```diff
@@ -13,6 +13,22 @@ public import Mathlib.Order.Interval.Set.Infinite
...
 
+/-!
+# Eta sets
+
+If `α` is a type with a `LinearOrder`, and `c` is some `Cardinal` in the same universe, then
+`IsEta c α` states that for any two subsets `X Y : Set α` of cardinality less than `c`, if every
+element of `X` is less than every element of `Y`, then there is some `(z : α)` greater than all
+elements of `X` and less than all elements of `Y`.
+
+Every linear order is vacuously `IsEta 0`. For `1 ≤ c ≤ ℵ₀`, the predicate `IsEta c` is equivalent
```
plp127, line 24: That's not right
```suggestion
Every linear order is vacuously `IsEta 0`. For `1 < c ≤ ℵ₀`, the predicate `IsEta c` is equivalent
```

vihdzp: oh oops I'll fix that later

PR #417, opened 2026-05-27
feat: endow `Simplicity` with the Scott-Hausdorff topology

vihdzp:
Can we add a `references.bib` file like Mathlib does?

- [x] depends on: #399


CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -62,20 +62,26 @@ def Simplicity : Type _ := SignExpansion
...
+open Order Set
 
 /-- The identity function between `SignExpansion` and `Simplicity`. -/
 def of : SignExpansion ≃ Simplicity := Equiv.refl _
 
 /-- The identity function between `Simplicity` and `SignExpansion`. -/
 def val : Simplicity ≃ SignExpansion := Equiv.refl _
 
+@[simp, grind =] theorem of_symm : of.symm = val := rfl
+@[simp, grind =] theorem val_symm : val.symm = of := rfl
```
plp127:
```suggestion
@[simp, grind =] theorem symm_of : of.symm = val := rfl
@[simp, grind =] theorem symm_val : val.symm = of := rfl
```

CombinatorialGames/SignExpansion/Simplicity.lean
```diff
@@ -136,31 +161,57 @@ theorem le_of_le_of_length_le {x y z : Simplicity} (hx : x ≤ z) (hy : y ≤ z)
...
+  obtain rfl | hxy := eq_or_ne x y
+  · simp
+  · obtain h | h := le_or_ge_of_le hx hy
+    · exact .inl <| h.lt_of_ne hxy
+    · exact .inr h
+
+theorem le_or_gt_of_le {x y z : Simplicity} (hx : x ≤ z) (hy : y ≤ z) : x ≤ y ∨ y < x :=
+  (lt_or_ge_of_le hy hx).symm
+
+theorem Iic_diff_Iic {x y : Simplicity} (h : x ≤ y) : Iic y \ Iic x = Ioc x y := by
```
plp127: This should be called with `sdiff` instead of `diff`.

CombinatorialGames/SignExpansion/Topology.lean
```diff
@@ -0,0 +1,191 @@
...
+
+public section
+
+namespace Simplicity
+open Order Set SignExpansion Topology
+
+instance : TopologicalSpace NatOrdinal := Preorder.topology _
+instance : OrderTopology NatOrdinal := ⟨rfl⟩
+instance : TopologicalSpace (WithTop NatOrdinal) := Preorder.topology _
+instance : OrderTopology (WithTop NatOrdinal) := ⟨rfl⟩
```
plp127: Can you move these to some file about `NatOrdinal` instead? Or would that increase imports? I don't want to be working with `NatOrdinal` and want a topology and import all the `NatOrdinal` files and not find the topology because they were actually defined in this file.

vihdzp: That would increase imports by quite a lot, but I can put them in a `NatOrdinal/Topology` file which might in the future also house specific theorems about it (directly ported from the theorems on `Ordinal`, of course).

PR #421, opened 2026-08-02
feat: `switch_equiv_zero_iff`

plp127:
Prove that `±x ≈ 0 ↔ ¬0 ≤ x`.

PR #422, opened 2026-08-04
feat: `switch_fuzzy_zero_iff` and `switch_fuzzy_self_iff`

plp127:
Characterize when `switch x` is confused to `0` and when it's confused to `x`. Follow-up to #421.

---

- [x] depends on: #424

### DIFF
--- a/CombinatorialGames/Game/Special.lean
+++ b/CombinatorialGames/Game/Special.lean
@@ -238,10 +238,9 @@ theorem switch_equiv_zero_iff {x : IGame} : ±x ≈ 0 ↔ ¬0 ≤ x := by
 alias ⟨_, switch_equiv_zero⟩ := switch_equiv_zero_iff
 
 theorem switch_fuzzy_zero_iff {x : IGame} : ±x ‖ 0 ↔ 0 ≤ x := by
-  refine Iff.trans ?_ switch_equiv_zero_iff.not_left
-  unfold AntisymmRel IncompRel
-  beta_reduce
-  rw [← IGame.zero_le_neg, neg_switch, and_self, and_self]
+  rw [← not_iff_not, ← switch_equiv_zero_iff]
+  exact not_fuzzy_iff_of_equiv_neg
+    (neg_switch x).symm.antisymmRel neg_zero.symm.antisymmRel
 
 alias ⟨_, switch_fuzzy_zero⟩ := switch_fuzzy_zero_iff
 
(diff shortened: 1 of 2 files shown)


CombinatorialGames/Game/Special.lean
```diff
@@ -237,5 +237,31 @@ theorem switch_equiv_zero_iff {x : IGame} : ±x ≈ 0 ↔ ¬0 ≤ x := by
 
 alias ⟨_, switch_equiv_zero⟩ := switch_equiv_zero_iff
 
+theorem switch_fuzzy_zero_iff {x : IGame} : ±x ‖ 0 ↔ 0 ≤ x := by
```
vihdzp, line 240: I think you can generalize this to `x ≈ y ↔ ¬ x ∥ y` whenever `x ≈ -x` and `y ≈ -y`. This generalization should also be useful for golfing some theorems on impartial games.

CombinatorialGames/Game/Special.lean
```diff
@@ -237,5 +237,31 @@ theorem switch_equiv_zero_iff {x : IGame} : ±x ≈ 0 ↔ ¬0 ≤ x := by
...
+  by_cases h0x : 0 ≤ x
+  · apply iff_of_true
+    · constructor
+      · apply left_lf
+        simp
+      · intro h
+        have hlt : 0 < ±x := h0x.trans_lt (lt_of_le_not_ge h (left_lf (by simp)))
+        apply hlt.not_gt
+        rw [← neg_switch, IGame.neg_lt_zero]
+        exact hlt
```
vihdzp: Likewise, these lines could be golfed if we had the theorem that `¬ x < y` whenever `x ≈ -x` and `y ≈ -y`.

CombinatorialGames/Game/Special.lean
plp127: I think this works more generally for two finite order elements (`x ≈ -x` is order dividing 2)

plp127: and, more generally it works for any ordered group

plp127: opened #424

CombinatorialGames/Game/Special.lean
```diff
@@ -237,5 +237,28 @@ theorem switch_equiv_zero_iff {x : IGame} : ±x ≈ 0 ↔ ¬0 ≤ x := by
...
+
+theorem switch_fuzzy_self_iff {x : IGame} : ±x ‖ x ↔ ¬x < 0 := by
+  by_cases h0x : 0 ≤ x
+  · apply iff_of_true
+    · constructor
+      · apply left_lf
+        simp
+      · intro h
+        absurd h0x.trans_lt (lt_of_le_not_ge h (left_lf (by simp)))
+        exact not_lt_of_equiv_neg neg_zero.symm.antisymmRel (neg_switch x).symm.antisymmRel
```
vihdzp, line 255: Huh, something I didn't notice about #424 is that we usually write `-x ≈ x` rather than `x ≈ -x`. Perhaps we should switch them around?

vihdzp, line 255:
```suggestion
  · refine iff_of_true ⟨left_lf ?_, fun h ↦ ?_⟩ h0x.not_gt
    · simp
    · absurd h0x.trans_lt (lt_of_le_not_ge h (left_lf (by simp)))
      exact not_lt_of_equiv_neg neg_zero.symm.antisymmRel (neg_switch x).symm.antisymmRel
```

plp127: perhaps?

vihdzp: I mean, the alternative is switching the entirety of the `Impartial` API to use `x ≈ -x`, which seems tedious and ugly.

plp127: I guess I can change these in the same PR then?

vihdzp (approved):

PR #425, opened 2026-08-05
fix: build is noisy

plp127:
Fix noisy build, and make CI fail on noisy build so this doesn't happen again.


vihdzp:
Thanks! I've been meaning to fix that for a while...

PR #426, opened 2026-08-05
feat: `birthday_intCast`

plp127:
Prove the theorem characterizing birthday of integers.


vihdzp (approved):
Huh, weird we didn't have this already!

PR #429, opened 2026-08-08
perf(Game/IGame): speedup

plp127:
Speedup the slowest declarations in `CombinatorialGames.Game.IGame`.

---

- [x] depends on: leanprover-community/mathlib4#42549


vihdzp:
Since this PR isn't critical, can we perhaps wait for the Mathlib PR to get in first?

Also, I think I wrote some of these `Set.forall_` lemmas in the `GameCmp` tactic file. Can you ensure they're not being duplicated there?

plp127:
> Also, I think I wrote some of these `Set.forall_` lemmas in the `GameCmp` tactic file. Can you ensure they're not being duplicated there?

indeed `Set.forall_mem_union` is in `CombinatorialGames.Game.Loopy.Basic` (but it's not `public`)

plp127:
The mathlib PR has been merged :)

vihdzp (approved):
Eh, I don't like these PRs all too much. But these proofs were already unreadable, this doesn't make it much worse.

PR #124, opened 2025-07-12
feat: dominated moves, reversible moves, and gift horses

plp127:

PR #430, opened 2026-08-11
feat: reverse all options of a game

plp127:
Define `unreverse (x : IGame) : IGame`, which recursively reverses all the options of `x` until they can't be reversed anymore, so that `unreverse x` is a game equivalent to `x` which has no reversible options.

---

Unfortunately I wan't able to use the theorems from #124, since those are only for reversing moves once, but this definition needs to be able to reverse possibly infinitely many times, so I can't even use transitivity since that only chains finitely many times.

### DIFF
--- a/CombinatorialGames/Game/Canonical.lean
+++ b/CombinatorialGames/Game/Canonical.lean
@@ -26,6 +26,33 @@ public noncomputable section
 
 namespace IGame
 
+def reverseSet (x : IGame) (p : Player) (z : IGame) : Set IGame :=
+  {g | g ∈ z.moves (-p) ∧ p.cases (g ≤ x) (x ≤ g)}
+
+instance (x : IGame.{u}) (p : Player) (z : IGame.{u}) : Small.{u} (reverseSet x p z) := by
+  unfold reverseSet
+  infer_instance
+
+-- false positive on `hg` which is referenced in the termination proof
+set_option linter.unusedVariables false in
+def unreverse1 (x : IGame) (p : Player) (z : IGame) : Set IGame :=
+  open scoped Classical in
+  if reverseSet x p z = ∅ then {z} else
+  ⋃ (g) (hg : g ∈ reverseSet x p z), unreverse1 x p g
+termination_by z
+decreasing_by exact .of_mem_moves hg.1
+
+instance (x : IGame.{u}) (p : Player) (z : IGame.{u}) : Small.{u} (unreverse1 x p z) := by
+  fun_induction unreverse1 x p z with
+  | case1 => infer_instance
+  | case2 z _ ih => exact @small_biUnion _ _ (reverseSet x p z) _ _ ih
+
+@[expose]
+def unreverse (x : IGame) : IGame :=
+  !{fun p => ⋃ z : x.moves p, unreverse1 x p (unreverse z)}
+termination_by x
+decreasing_by igame_wf
+
 /-- Undominating a game. This returns garbage values on non-short games -/
 def undominate (x : IGame) : IGame :=
   !{{y ∈ Set.range fun z : xᴸ ↦ undominate z | ∀ z ∈ xᴸ, ¬y < z} |
--- a/CombinatorialGames/Game/Canonical.lean
+++ b/CombinatorialGames/Game/Canonical.lean
@@ -56,6 +56,11 @@ instance (x : IGame.{u}) (p : Player) (z : IGame.{u}) : Small.{u} (reverseSet x 
   unfold reverseSet
   infer_instance
 
+theorem neg_reverseSet (x : IGame) (p : Player) (z : IGame) :
+    -reverseSet x p z = reverseSet (-x) (-p) (-z) := by
+  unfold reverseSet
+  cases p <;> simp [Set.ext_iff, IGame.neg_le]
+
 -- false positive on `hg` which is referenced in the termination proof
 set_option linter.unusedVariables false in
 def unreverse1 (x : IGame) (p : Player) (z : IGame) : Set IGame :=
@@ -71,6 +76,21 @@ instance (x : IGame.{u}) (p : Player) (z : IGame.{u}) : Small.{u} (unreverse1 x 
   | case2 z _ ih =>
     exact @small_biUnion _ _ (reverseSet x p z) _ _ fun g hg =>
       @small_biUnion _ _ (g.moves p) _ _ (ih g hg)
+
+theorem neg_unreverse1 (x : IGame) (p : Player) (z : IGame) :
+    -unreverse1 x p z = unreverse1 (-x) (-p) (-z) := by
+  fun_induction unreverse1 x p z with
+  | case1 z hx =>
+    unfold unreverse1
+    rw [← neg_reverseSet, hx]
+    simp
+  | case2 z hx ih =>
+    rw [unreverse1, ← neg_reverseSet]
+    simp_rw [Set.neg_eq_empty, if_neg hx]
+    simp only [← Set.image_neg_eq_neg, Set.image_iUnion,
+      Set.biUnion_image, moves_neg, neg_neg]
+    refine Set.iUnion₂_congr fun g hg => Set.iUnion₂_congr fun g' hg' => ?_
+    rw [Set.image_neg_eq_neg, ih g hg g' hg']
 
 theorem lf_of_mem_reverseSet_of_mem_unreverse1
     (x : IGame) (p : Player) (z : IGame) {g g' c : IGame}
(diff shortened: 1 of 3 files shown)


CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,205 @@ games through undominating and unreversing games.
...
+-- mathlib PR #42549
+theorem forall_mem_biUnion {p : α → Prop} {f : ι → Set α} {q : ι → Prop} :
+    (∀ x ∈ ⋃ (i : ι) (_ : q i), f i, p x) ↔ (∀ i, q i → ∀ x ∈ f i, p x) :=
+  forall_mem_iUnion₂
+
+end ForMathlib
 
 namespace IGame
 
+def reverseSet (x : IGame) (p : Player) (z : IGame) : Set IGame :=
```
vihdzp: Docstring? Even if this is private, it'd be appreciated.

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,205 @@ games through undominating and unreversing games.
...
+  cases p <;> simp [Set.ext_iff, IGame.neg_le]
+
+theorem reverseSet_congr_left {x y : IGame} (hxy : x ≈ y) (p : Player) (z : IGame) :
+    reverseSet x p z = reverseSet y p z := by
+  unfold reverseSet
+  cases p <;> simp [hxy.le_congr_left, hxy.le_congr_right]
+
+-- false positive on `hg` which is referenced in the termination proof
+set_option linter.unusedVariables false in
+def unreverse1 (x : IGame) (p : Player) (z : IGame) : Set IGame :=
```
vihdzp, line 84: Same here

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,205 @@ games through undominating and unreversing games.
...
+  rw [ofSets_inj, rightMoves_ofSets, and_iff_left rfl,
+    Player.neg_right]
+  simp_rw [Set.iUnion_coe_set, ← Set.iSup_eq_iUnion]
+  apply (Equiv.neg IGame).iSup_congr
+  simp [neg_eq]
+
+public section
+
+/-- Recursively repeatedly bypass all reversible options from a game `x`,
+so that `unreverse x` hereditarily has no reversible options. -/
```
vihdzp, line 229: Can you add lemmas which prove the characterization mentioned in the docstring?

vihdzp (commented):
Sorry for the backlog, I'll try and review properly tonight.

CombinatorialGames/Game/Canonical.lean
plp127: We'd need a predicate for hereditarily having no reversible options. I've proved that it's equivalent to the original game.

vihdzp: Is it true that any option of `unreverse x` equals its own `unreverse`? If so, that plus the fact that `unreverse x` is nonreversible would imply this, at least tacitly. We should definitely have the latter, either way.

CombinatorialGames/Game/Canonical.lean
vihdzp: Would it be worth to make this public? Then we could show `reverseSet (unreverse x) p z = ∅` for `z ∈ x.moves p`.

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,212 @@ games through undominating and unreversing games.
...
+end ForMathlib
 
 namespace IGame
 
+/-- The set of `-p`-moves of `z` which reverse `z` with respect to a `p`-move from `x`.
+That is, if player `p` moves from `x` to `z`, then `reverseSet x p z` is the set of
+moves `-p` could make as a response that reverse the move from `x` to `z`.
+Note that `z` is not necessarily a `p`-option of `x`. -/
+def reverseSet (x : IGame) (p : Player) (z : IGame) : Set IGame :=
+  {g | g ∈ z.moves (-p) ∧ p.cases (g ≤ x) (x ≤ g)}
```
vihdzp, line 58: Starting to think that maybe a `Player.LE` predicate would be good to have after all. Writing all the PR seems daunting, but maybe we just add what we need at a given moment?

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,212 @@ games through undominating and unreversing games.
...
+/-- Repeatedly reverse the move `z` with respect to a `p`-move from `x`.
+Treating `z` as a `p`-option of `x`, bypass it if it is reversible, and
+then recursively reverse all the resulting games. -/
+def unreverse1 (x : IGame) (p : Player) (z : IGame) : Set IGame :=
+  open scoped Classical in
+  if reverseSet x p z = ∅ then {z} else
+  ⋃ (g) (hg : g ∈ reverseSet x p z) (g') (hg' : g' ∈ g.moves p), unreverse1 x p g'
+termination_by z
+decreasing_by exact .trans (.of_mem_moves hg') (.of_mem_moves hg.1)
+
```
vihdzp, line 90: Maybe it'd be worth adding `unreverse1_of_empty` and `unreverse1_of_nonempty` lemmas so that you don't have to constantly unfold the definition?

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,212 @@ games through undominating and unreversing games.
...
+    unreverse1 x p z = unreverse1 y p z := by
+  fun_induction unreverse1 x p z with
+  | case1 z hx => rw [unreverse1, ← reverseSet_congr_left hxy, if_pos hx]
+  | case2 z hx ih =>
+    rw [unreverse1, ← reverseSet_congr_left hxy, if_neg hx]
+    refine Set.iUnion₂_congr fun g hg => Set.iUnion₂_congr fun g' hg' => ?_
+    exact ih g hg g' hg'
+
+theorem lf_of_mem_reverseSet_of_mem_unreverse1
+    (x : IGame) (p : Player) (z : IGame) {g g' c : IGame}
```
vihdzp: Do `x`, `p`, `z` have to be explicit?

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,212 @@ games through undominating and unreversing games.
...
+    cases p with
+    | left => exact fun h => left_lf hg' (hg.2.trans h)
+    | right => exact fun h => lf_right hg' (h.trans hg.2)
+  · rw [if_neg hx] at hc
+    simp_rw [Set.mem_iUnion] at hc
+    obtain ⟨g'', hg'', g''', hg''', hc⟩ := hc
+    exact ih g' (.trans (.of_mem_moves hg') (.of_mem_moves hg.1)) hg'' hg''' hc
+
+theorem lf_of_mem_moves_of_mem_unreverse1
+    (x : IGame) (p : Player) (z : IGame) {g : IGame}
```
vihdzp: Same here

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,212 @@ games through undominating and unreversing games.
...
+      exact ⟨z, hz, ih left z hz⟩
+    · rw [moves_ofSets, Set.forall_mem_image]
+      intro z hz
+      exact ⟨z, hz, ih right z hz⟩
+    · intro z hz
+      rw [moves_ofSets, Set.exists_mem_image]
+      exact ⟨z, hz, ih left z hz⟩
+    · intro z hz
+      rw [moves_ofSets, Set.exists_mem_image]
+      exact ⟨z, hz, ih right z hz⟩
```
vihdzp:
```suggestion
    apply equiv_of_exists <;> simpa using fun z hz ↦ ⟨z, hz, ih _ z hz⟩
```

vihdzp:
Where does this leave #124? My major complaint there was that the theorems seemed hard to apply, and that seems to have been true...

plp127:
> Where does this leave #124? My major complaint there was that the theorems seemed hard to apply, and that seems to have been true...

The lemmas are still probably useful to have, but I don't see how to get them to apply in this case where bypassing one game could generate infinitely many options, the `n`-th one having to be bypassed `n` times, and suddenly you have an infinite amount of bypassing all coming from just one option. How would I state such a lemma?

CombinatorialGames/Game/Canonical.lean
```diff
@@ -22,10 +22,310 @@ games through undominating and unreversing games.
...
+public theorem lf_of_reverseSet_eq_empty_of_mem_moves {x : IGame} {p : Player} {z : IGame}
+    (hx : reverseSet x p z = ∅) {g : IGame} (hg : g ∈ z.moves (-p)) :
+    ¬p.cases (g ≤ x) (x ≤ g) :=
+  fun h => Set.eq_empty_iff_forall_notMem.1 hx g ⟨hg, h⟩
+
+-- false positive on `hg` which is referenced in the termination proof
+set_option linter.unusedVariables false in
+/-- Repeatedly reverse the move `z` with respect to a `p`-move from `x`.
+Treating `z` as a `p`-option of `x`, bypass it if it is reversible, and
+then recursively reverse all the resulting games. -/
```
vihdzp:
```suggestion
/-- Repeatedly unreverse the move `z` with respect to a `p`-move from `x`.
Treating `z` as a `p`-option of `x`, bypass it if it is reversible, and
then recursively unreverse all the resulting games. -/
```

PR #438, opened 2026-08-23
feat: moves of `nsmul`

plp127:
Show that `((n + 1) • x).moves p = (n • x + ·) '' x.moves p`.

### DIFF
--- a/CombinatorialGames/Game/Basic.lean
+++ b/CombinatorialGames/Game/Basic.lean
@@ -103,8 +103,8 @@ instance : AddCommGroupWithOne Game where
   add_comm := by rintro ⟨x⟩ ⟨y⟩; exact congr(mk $(add_comm _ _))
   add_assoc := by rintro ⟨x⟩ ⟨y⟩ ⟨z⟩; exact congr(mk $(add_assoc _ _ _))
   neg_add_cancel := by rintro ⟨a⟩; exact mk_eq (neg_add_equiv _)
-  nsmul_zero := by rintro ⟨a⟩; exact congr(mk $(zero_nsmul _))
-  nsmul_succ := by rintro n ⟨a⟩; exact congr(mk $(succ_nsmul _ _))
+  nsmul_zero := by rintro ⟨a⟩; rfl
+  nsmul_succ := by rintro n ⟨a⟩; rfl
   zsmul_zero' := by rintro ⟨a⟩; rfl
   zsmul_succ' := by rintro n ⟨a⟩; rfl
   zsmul_neg' := by rintro n ⟨a⟩; rfl
--- a/CombinatorialGames/Game/Basic.lean
+++ b/CombinatorialGames/Game/Basic.lean
@@ -103,8 +103,8 @@ instance : AddCommGroupWithOne Game where
   add_comm := by rintro ⟨x⟩ ⟨y⟩; exact congr(mk $(add_comm _ _))
   add_assoc := by rintro ⟨x⟩ ⟨y⟩ ⟨z⟩; exact congr(mk $(add_assoc _ _ _))
   neg_add_cancel := by rintro ⟨a⟩; exact mk_eq (neg_add_equiv _)
-  nsmul_zero := by rintro ⟨a⟩; rfl
-  nsmul_succ := by rintro n ⟨a⟩; rfl
+  nsmul_zero := by rintro ⟨a⟩; exact congr(mk $(zero_nsmul _))
+  nsmul_succ := by rintro n ⟨a⟩; exact congr(mk $(succ_nsmul _ _))
   zsmul_zero' := by rintro ⟨a⟩; rfl
   zsmul_succ' := by rintro n ⟨a⟩; rfl
   zsmul_neg' := by rintro n ⟨a⟩; rfl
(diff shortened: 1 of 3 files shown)


CombinatorialGames/Game/IGame.lean
```diff
@@ -774,12 +774,36 @@ private theorem add_assoc' (x y z : IGame) : x + y + z = x + (y + z) := by
 termination_by (x, y, z)
 decreasing_by igame_wf
 
+private def nsmul' (n : Nat) (x : IGame) : IGame :=
```
vihdzp: Why the custom `nsmul` def if it's just going to end up equal to the default implementation?

plp127: wdym? every possible `nsmul` is always gonna equal to default!

vihdzp: Yeah, so the question is what do we get by reimplementing it?

plp127: we get a really easy time proving `moves_succ_nsmul` (look it's just two lines!)

vihdzp: Seems to me that we're just moving around what's easy and what's hard. Is it really that much difficult to prove this theorem without using the def-eq? Then you'd save like 10 lines in proving that this `nsmul` satisfies the same properties as the default implementation.

plp127: well then I'm guessing I would also lose 10 lines proving the theorem without the defeq so it's +10 -10 and nothing has changed

CombinatorialGames/Game/IGame.lean
```diff
@@ -791,6 +815,11 @@ theorem moves_sub (p : Player) (x y : IGame) :
     (x - y).moves p = (· - y) '' x.moves p ∪ (x + ·) '' (-y.moves (-p)) := by
   simp [sub_eq_add_neg]
 
+theorem moves_succ_nsmul (p : Player) (n : Nat) (x : IGame) :
+    ((n + 1) • x).moves p = (n • x + ·) '' x.moves p := by
+  change (nsmul' (n + 1) x).moves p = (nsmul' n x + ·) '' x.moves p
+  rw [nsmul', moves_ofSets]
```
vihdzp: Here's a 7-line proof that doesn't require adjusting def-eqs.
```suggestion
theorem moves_succ_nsmul (p : Player) (n : Nat) (x : IGame) :
    ((n + 1) • x).moves p = (n • x + ·) '' x.moves p := by
  induction n with
  | zero => simp
  | succ n IH =>
    rw [succ_nsmul, moves_add, IH, union_eq_right]
    simp_rw [image_subset_iff, subset_def, mem_preimage, mem_image]
    refine fun y hy ↦ ⟨y, hy, ?_⟩
    rw [succ_nsmul, add_right_comm]
```

CombinatorialGames/Game/IGame.lean
```diff
@@ -775,10 +775,10 @@ termination_by (x, y, z)
 decreasing_by igame_wf
 
 instance : AddCommMonoid IGame where
-  add_zero := private add_zero'
-  zero_add _ := private add_comm' .. ▸ add_zero' _
-  add_comm := private add_comm'
-  add_assoc := private add_assoc'
+  add_zero := by exact add_zero'
```
vihdzp, line 778: What's the advantage of `by exact` over `private`?

CombinatorialGames/Game/Basic.lean
```diff
@@ -103,8 +103,8 @@ instance : AddCommGroupWithOne Game where
   add_comm := by rintro ⟨x⟩ ⟨y⟩; exact congr(mk $(add_comm _ _))
   add_assoc := by rintro ⟨x⟩ ⟨y⟩ ⟨z⟩; exact congr(mk $(add_assoc _ _ _))
   neg_add_cancel := by rintro ⟨a⟩; exact mk_eq (neg_add_equiv _)
-  nsmul_zero := by rintro ⟨a⟩; rfl
-  nsmul_succ := by rintro n ⟨a⟩; rfl
+  nsmul_zero := by rintro ⟨a⟩; exact congr(mk $(zero_nsmul _))
```
vihdzp: What's wrong with these `rfl`s? Are they abusive somehow?

plp127: This was carried over from when I had changed the defeqs of `nsmul` and had to fix this proof which broke. If you think it's fine I can also revert this change.

CombinatorialGames/Game/IGame.lean
plp127: This was carried over from when I touched this instance when changing the defeqs of `nsmul`. I think `private` and `by exact` are functionally mostly the same but I prefer using `by exact` over private because it additionally communicates that this is a proof and also the auxiliary definitions it generates are `._proof_n` `theorem`s instead of `._private_n` `def`s.

CombinatorialGames/Game/Basic.lean
vihdzp: I think it's fine to keep as `rfl`, as long as the linters don't start screaming at us.

vihdzp (approved):

PR #440, opened 2026-09-07
feat: `LinearOrderedCommMonoidWithZero NatOrdinal`

vihdzp:


plp127:
What do we need this for?

vihdzp:
I found it within the AI formalization of Conway's refinement conjecture... so presumably it's used for something?
Even if not it's probably a nice instance to have.

plp127 (approved):

PR #443, opened 2026-09-09
feat: natural products and embeddings

vihdzp:


CombinatorialGames/NatOrdinal/OrderType.lean
```diff
@@ -0,0 +1,364 @@
...
+import Mathlib.Order.GameAdd
+
+/-!
+# Maximal order types of embeddings
+
+This file is devoted to two theorems, which also serve as characterizations of the natural sum and
+product. Let `α`, `β`, `γ` be well-orders.
+
+- Given the existence of a monotone, surjective map `f : α ⊕ β → γ`, the order type of `γ` is at
+  most the natural sum of the order types of `α` and `β`, and the inequality is strict.
```
plp127, line 18: The inequality is not a *strict inequality* (it does not have a `<` in it). Maybe instead you can say that "the equality case is achieved"?

PR #444, opened 2026-09-11
feat: update Mathlib

vihdzp:
https://github.com/leanprover-community/mathlib4/pull/42351 just got merged, so I felt it pertinent to make a mid-version update.


CombinatorialGames/Game/Specific/Nim.lean
```diff
@@ -153,7 +153,7 @@ theorem _root_.Game.birthday_nim (o : Nimber) : Game.birthday (.mk (nim o)) = .o
   · exact hu hy hxy
   have hyo := lf_right_of_le hxo.ge hy
   replace hy := Subposition.of_mem_moves hy
-  induction y using IsWellFounded.induction Subposition with | ind y ihy
+  induction y using wellFounded_subposition.induction with | h y ihy
```
plp127, line 156: This seems like a regression to me, the case name `ind` is replaced by the less descriptive `h`. Unfortunately the induction theorem is in core...

vihdzp: I agree with you. Making core PRs is always intimidating, though.

CombinatorialGames/Game/IGame.lean
```diff
@@ -198,7 +198,7 @@ instance : IsTrans _ Subposition := inferInstanceAs (IsTrans _ (Relation.TransGe
 instance small_setOf_subposition (x : IGame.{u}) : Small.{u} {y | Subposition y x} :=
   small_transGen' _ x
 
-theorem subposition_wf : WellFounded Subposition := by
+instance wellFounded_subposition : WellFounded Subposition := by
```
plp127, line 201: Why have we changed the name of this? I think `subposition_wf` was a fine name.

vihdzp: I thought the `_wf` naming convention was a bit ad hoc. Should I revert these renames?

plp127: It probably doesn't matter, as long as we're consistent.

PR #447, opened 2026-09-12
feat: `Algebra ℝ Surreal`

vihdzp:
Used in the refinement theorem formalization and seems useful regardless; I golfed down the proofs in that repository.

### DIFF
--- a/CombinatorialGames/Surreal/Real.lean
+++ b/CombinatorialGames/Surreal/Real.lean
@@ -626,6 +626,19 @@ def toSurrealRingHom : ℝ →+*o Surreal where
   map_mul' := toSurreal_mul
   monotone' := toSurrealEmbedding.monotone
 
+instance : Algebra ℝ Surreal :=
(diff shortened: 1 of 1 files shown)


CombinatorialGames/Surreal/Real.lean
```diff
@@ -626,6 +626,19 @@ def toSurrealRingHom : ℝ →+*o Surreal where
   map_mul' := toSurreal_mul
   monotone' := toSurrealEmbedding.monotone
 
+instance : Algebra ℝ Surreal :=
```
plp127, line 629: If we're adding this instance, can we make the canonical form of a real number to be `algebraMap ℝ Surreal x`?

vihdzp: That's quite a mouthful, is it really better than just having the coercion?

plp127: The benefit is you get all the theorems about `algebraMap` for free

PR #452, opened 2026-09-15
feat: update docstrings, add theorems to special games

LLaurance:
I promise that in updating the docstrings this time I did not go over the line length.

The upstart equality is in Ch. 7 of Of Numbers and Games and disguised in Ch. 3 of Mathematical Go.


CombinatorialGames/Game/Special.lean
```diff
@@ -134,6 +142,8 @@ theorem leftMoves_tiny (x : IGame) : (⧾x)ᴸ = {0} :=
 theorem rightMoves_tiny (x : IGame) : (⧾x)ᴿ = {!{{0} | {-x}}} :=
   rightMoves_ofSets ..
 
+theorem tiny_zero_eq_up : ⧾0 = ↑ := by aesop
```
plp127: Why `aesop`? Does `game_cmp` not work here?

LLaurance: It works if the `=` is `≈` instead.

plp127: Is this something it should be handling? Is this in scope for `game_cmp`?

LLaurance: Unfortunately I do not know the answer to this question 🥲

vihdzp: I don't think it is. `game_cmp` only compares games, it isn't designed to prove equalities.

What is true is that this lemma should be tagged as `game_cmp`, and maybe `simp` too.

CombinatorialGames/Game/Special.lean
```diff
@@ -182,6 +192,8 @@ theorem leftMoves_miny (x : IGame) : (⧿x)ᴸ = {!{{x} | {0}}} :=
 theorem rightMoves_miny (x : IGame) : (⧿x)ᴿ = {0} :=
   rightMoves_ofSets ..
 
+theorem miny_zero_eq_down : ⧿0 = ↓ := by aesop
```
vihdzp: Same here.

vihdzp (commented):
Other than the one suggestion, this LGTM.

CombinatorialGames/Game/Special.lean
LLaurance: I was hesitant to tag `simp` since there is a considerable difference between tinies/minies and ups/downs (Section 2.5 Mathematical Go), but on a second read it is stated up/down is preferable to tiny/miny so I will tag both.

CombinatorialGames/Game/Special.lean
LLaurance: Resolved https://github.com/vihdzp/combinatorial-games/pull/452/changes/bccbcc2371ac89e718157b8ad924ccdbbfb73402

CombinatorialGames/Game/Special.lean
LLaurance: Resolved https://github.com/vihdzp/combinatorial-games/pull/452/changes/bccbcc2371ac89e718157b8ad924ccdbbfb73402

vihdzp (approved):

